PR #3, opened 2022-02-01
Dapps-Staking evm precompile

Maar-io:
**extrinsics:**
- [x] register()
- [x] bond_and_stake()
- [x] unbond_and_unstake()
- [x] withdraw_unbonded()
- [x] claim()

**storage getters()**
- [x] read_current_era
- [x] read_era_reward
- [x] read_era_staked
- [x] read_staked_amount
- [x] read_contract_era_stake

**Other To-dos**
- [x] design Solidity interface

**Check list**
- [x] adds new feature
- [ ] updated spec version
- [ ] documentation changes

**Adds**
- precompiled contract for dApps Staking pallet

**Fixes**
- None

**Changes**
- also changes some storage scope in the dapps-staking-pallet


precompiles/dapps-staking/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+sha3 = "0.9"
+sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13",  default-features = false }
+sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13" }
+scale-info = { version = "1.0", default-features = false, features = ["derive"] }
+
+[features]
+default = ["std"]
+std = [
+	"codec/std",
+    "scale-info/std",
```
Dinonard, line 40: Formatting

precompiles/dapps-staking/DappsStaking.sol
```diff
@@ -0,0 +1,52 @@
...
+interface DappsStaking {
+
+    // Storage getters
+
+    /// @dev Read current era.
+    /// @return The current era
+    function read_current_era() external view returns (uint256);
+
+    /// @dev Read unbonding period constant.
+    /// @return The unbonding period
```
Dinonard: in eras

precompiles/dapps-staking/DappsStaking.sol
```diff
@@ -0,0 +1,52 @@
...
+    /// @return Total staked amount for the given era
+    function read_era_staked(uint32 era) external view returns (uint128);
+
+    /// @dev Read Staked amount for the staker
+    /// @return Staked amount for the staker
+    function read_staked_amount(address staker) external view returns (uint128);
+
+    /// @dev Read the amount staked on contract in the given era
+    /// @return The amount staked on contract in the given era
+    function read_contract_era_stake(address contract_id, uint32 era) external view returns (uint128);
```
Dinonard: I'd recommend you change this to just accept contract_id.
Eras don't mean that much outside of our dapps-staking.

precompiles/dapps-staking/DappsStaking.sol
```diff
@@ -0,0 +1,52 @@
...
+    /// @dev Stake provided amount on the contract.
+    function bond_and_stake(address, uint128) external;
+
+    /// @dev Start unbonding process and unstake balance from the contract.
+    function unbond_and_unstake(address, uint128) external;
+
+    /// @dev Withdraw all funds that have completed the unbonding process.
+    function withdraw_unbonded() external;
+
+    /// @dev Claim contract's rewards.
```
Dinonard, line 50: Could you add a comment here that this should be used with care since it will be deprecated soon?

Dinonard (commented):
I didn't go through all the code again, just commented some parts in .sol

precompiles/dapps-staking/DappsStaking.sol
Maar-io: @Dinonard not sure what would be use of this method if era is removed. The procompile is calling `Pallet::<R>::staking_info(&contract_id, era)` 
Without era there is not much to call. Do you mean we then call latest stored/changed staking info?

precompiles/dapps-staking/DappsStaking.sol
Maar-io: done

precompiles/dapps-staking/DappsStaking.sol
Dinonard: Yes, we'd just return the latest value.

The reasoning behind is that I'm not yet sure how our storage schema will evolve in the future.
Consider the current way of storing this piece of information - we have a _sparse array_ in a sense. This storage will grow forever if we never clean old history - each era in which contract stake is modified, a new entry is created in the DB.

Now, if we decide to clean the history (and I'm pretty sure we'll have to do that because unbounded DB growth means problems), you would be returning false information from this method when `era` is very old.

Imo, adding method that returns **latest** staked value should be enough - in the future, when pallet matures more, we can extend this interface with a new method that takes eras.

precompiles/dapps-staking/DappsStaking.sol
```diff
@@ -0,0 +1,52 @@
...
+    /// @return Total network reward for the given era
+    function read_era_reward(uint32 era) external view returns (uint128);
+
+    /// @dev Read Total staked amount for the given era
+    /// @return Total staked amount for the given era
+    function read_era_staked(uint32 era) external view returns (uint128);
+
+    /// @dev Read Staked amount for the staker
+    /// @return Staked amount for the staker
+    function read_staked_amount(address staker) external view returns (uint128);
```
akru, line 29: Probably, using H160 everywhere instead of contract id will make troubles in the future, but in other case we could just overload same method with bytes32 AccountId when needed.

precompiles/dapps-staking/DappsStaking.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    /// @dev Read the amount staked on contract in the given era
+    /// @return The amount staked on contract in the given era
+    function read_contract_era_stake(address contract_id, uint32 era) external view returns (uint128);
+
+
+    // Extrinsic calls
+
+    /// @dev Register provided contract.
+    function register(address) external;
```
hoonsubin, line 39: @Maar-io do we have a H160 equivalent for the developer preapproval? How can H160 accounts submit their dapps?

Dinonard (approved):
LGTM

precompiles/dapps-staking/DappsStaking.sol
Maar-io: implemented without era

precompiles/dapps-staking/DappsStaking.sol
Maar-io: @hoonsubin `address` is H160

precompiles/dapps-staking/DappsStaking.sol
Maar-io: @akru do you mean to add following
  ` function read_staked_amount(uint256 staker) external view returns (uint128);`
We actually need it for `astarbase` since we need to verify 32bit public polkadot address

precompiles/dapps-staking/src/lib.rs
```diff
@@ -253,6 +253,7 @@ where
             [0x77, 0xa0, 0xfe, 0x02] => Self::withdraw_unbonded()?,
             [0xc1, 0x3f, 0x4a, 0xf7] => Self::claim(input)?,
             _ => {
+                println!("selector{:x?}", selector);
```
Dinonard: How does compiler work here? :scream:

Maar-io: thanks for finding this!!

Dinonard: Sure, but do you know why git action passed here?
Does it work locally for you?

PR #4, opened 2022-02-01
fixed era length: feature, test, OnRuntimeUpgrade

jakehemmerle:
This contains bugfix for fixing era length, related tests, and OnRuntimeUpgrade hooks. Documentation for configuring it for each network is included

**Check list**
- [x] tests

### DIFF
--- a/frame/dapps-staking/Cargo.toml
+++ b/frame/dapps-staking/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-dapps-staking"
-version = "3.0.4"
+version = "3.1.0"
 authors = ["Stake Technologies <devops@stake.co.jp>"]
 edition = "2018"
 homepage = "https://astar.network/"
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -163,6 +163,11 @@ pub mod pallet {
     #[pallet::getter(fn force_era)]
     pub type ForceEra<T> = StorageValue<_, Forcing, ValueQuery, ForceEraOnEmpty>;
 
+    /// Stores the block number of when the next era starts
+    #[pallet::storage]
+    #[pallet::getter(fn next_era_starting_block)]
+    pub type NextEraStartingBlock<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
+
     /// Registered developer accounts points to coresponding contract
     #[pallet::storage]
     #[pallet::getter(fn registered_contract)]
@@ -322,16 +327,16 @@ pub mod pallet {
             }
 
             let force_new_era = Self::force_era().eq(&Forcing::ForceNew);
-            let blocks_per_era = T::BlockPerEra::get();
             let previous_era = Self::current_era();
+            let next_era_starting_block = Self::next_era_starting_block();
 
             // Value is compared to 1 since genesis block is ignored
-            if now % blocks_per_era == BlockNumberFor::<T>::from(1u32)
-                || force_new_era
-                || previous_era.is_zero()
-            {
+            if now >= next_era_starting_block || force_new_era || previous_era.is_zero() {
+                let blocks_per_era = T::BlockPerEra::get();
                 let next_era = previous_era + 1;
                 CurrentEra::<T>::put(next_era);
+
+                NextEraStartingBlock::<T>::put(now + blocks_per_era);
 
                 let reward = BlockRewardAccumulator::<T>::take();
                 Self::reward_balance_snapshoot(previous_era, reward);
--- a/frame/dapps-staking/src/tests.rs
+++ b/frame/dapps-staking/src/tests.rs
@@ -92,6 +92,93 @@ fn on_initialize_is_ok() {
 }
 
 #[test]
+fn new_era_length_is_always_blocks_per_era() {
+    ExternalityBuilder::build().execute_with(|| {
+        initialize_first_block();
+        let blocks_per_era = mock::BLOCKS_PER_ERA;
+
+        // go to beginning of an era
+        advance_to_era(mock::DappsStaking::current_era() + 1);
+
+        // record era number and block number
+        let start_era = mock::DappsStaking::current_era();
+        let starting_block_number = System::block_number();
+
+        // go to next era
+        advance_to_era(mock::DappsStaking::current_era() + 1);
+        let ending_block_number = System::block_number();
+
+        // make sure block number difference is is blocks_per_era
+        assert_eq!(mock::DappsStaking::current_era(), start_era + 1);
+        assert_eq!(ending_block_number - starting_block_number, blocks_per_era);
+    })
+}
+
+#[test]
+fn new_era_is_handled_with_maintenance_mode() {
+    ExternalityBuilder::build().execute_with(|| {
+        initialize_first_block();
+
+        // enable maintenance mode
+        assert_ok!(DappsStaking::maintenance_mode(Origin::root(), true));
+        assert!(PalletDisabled::<TestRuntime>::exists());
+        System::assert_last_event(mock::Event::DappsStaking(Event::MaintenanceMode(true)));
+
+        // advance 9 blocks or 3 era lengths (advance_to_era() doesn't work in maintenance mode)
+        run_for_blocks(mock::BLOCKS_PER_ERA * 3);
+
+        // verify that `current block > NextEraStartingBlock` but era hasn't changed
+        assert!(System::block_number() > DappsStaking::next_era_starting_block());
+        assert_eq!(DappsStaking::current_era(), 1);
+
+        // disable maintenance mode
+        assert_ok!(DappsStaking::maintenance_mode(Origin::root(), false));
+        System::assert_last_event(mock::Event::DappsStaking(Event::MaintenanceMode(false)));
+
+        // advance one era
+        run_for_blocks(mock::BLOCKS_PER_ERA);
+
+        // verify we're at block 14
+        assert_eq!(System::block_number(), (4 * mock::BLOCKS_PER_ERA) + 2); // 2 from initialization, advanced 4 eras worth of blocks
+
+        // verify era was updated and NextEraStartingBlock is 15
+        assert_eq!(DappsStaking::current_era(), 2);
+        assert_eq!(
+            DappsStaking::next_era_starting_block(),
+            (5 * mock::BLOCKS_PER_ERA)
+        );
+    })
+}
+
+#[test]
+fn new_forced_era_length_is_always_blocks_per_era() {
+    ExternalityBuilder::build().execute_with(|| {
+        initialize_first_block();
+        let blocks_per_era = mock::BLOCKS_PER_ERA;
+
+        // go to beginning of an era
+        advance_to_era(mock::DappsStaking::current_era() + 1);
+
+        // go to middle of era
+        run_for_blocks(1); // can be any number between 0 and blocks_per_era
+
+        // force new era
+        <ForceEra<TestRuntime>>::put(Forcing::ForceNew);
+        run_for_blocks(1); // calls on_initialize()
+
+        // note the start block number of new (forced) era
+        let start_block_number = System::block_number();
+
+        // go to start of next era
+        advance_to_era(mock::DappsStaking::current_era() + 1);
+
+        // show the length of the forced era is equal to blocks_per_era
+        let end_block_number = System::block_number();
+        assert_eq!(end_block_number - start_block_number, blocks_per_era);
+    })
+}
+
+#[test]
 fn new_era_is_ok() {
     ExternalityBuilder::build().execute_with(|| {
         // set initial era index
--- a/precompiles/dapps-staking/Cargo.toml
+++ b/precompiles/dapps-staking/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-precompile-dapps-staking"
-version = "3.0.4"
+version = "3.1.0"
 authors = ["Stake Technologies <devops@stake.co.jp>"]
 edition = "2018"
 license = "Apache-2.0"


frame/dapps-staking/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-dapps-staking"
-version = "2.0.0"
+version = "2.1.0"
```
jakehemmerle: @Dinonard I should bump this yes?

frame/dapps-staking/src/migrations.rs
```diff
@@ -353,3 +353,31 @@ pub mod v2 {
         Ok(())
     }
 }
+
+pub mod v3 {
```
Dinonard: This isn't v3 so no need to have it here. Also, it isn't really a migration.

I'd recommend you just do this in an struct implementing the OnRuntimeUpgrade trait.

frame/dapps-staking/src/migrations.rs
```diff
@@ -353,3 +353,31 @@ pub mod v2 {
...
+        let current_era = CurrentEra::<T>::get();
+
+        // for local upgrade testing, as prev runtime version INCLUDES dapps staking (block 1 starts era 1)
+        // let next_era_starting_block =  ((current_era + 1) * blocks_per_era) - blocks_per_era;
+
+        // for shibuya/shiden/astar upgrade (or any network where dapps staking was NOT included in genesis)
+        let next_era_starting_block = dapps_staking_block_offset + ((current_era + 1) * blocks_per_era);
+
+        NextEraStartingBlock::<T>::put(next_era_starting_block.saturated_into::<T::BlockNumber>());
+        T::DbWeight::get().reads(2) + T::DbWeight::get().writes(1)
```
Dinonard: You only have one read :)

frame/dapps-staking/src/tests.rs
```diff
@@ -1292,6 +1292,58 @@ fn new_era_forcing() {
...
+        // note the start block number of new (forced) era
+        let start_block_number = System::block_number();
+        
+        // go to start of next era
+        advance_to_era(mock::DappsStaking::current_era() + 1);
+
+        // show the length of the forced era is equal to blocks_per_era
+        let end_block_number = System::block_number();
+        assert_eq!(end_block_number - start_block_number, blocks_per_era);
+    })
```
Dinonard: Good tests!

Dinonard (commented):
Please make sure to run `cargo fmt` before you push.

Good progress so far!

How about benchmarking the `on_initialize` (https://github.com/AstarNetwork/Astar/issues/552)?

Regarding the commit in Astar repo, please rebase it and also run `cargo fmt`.
I'll also look at how this and that commit work together.

jakehemmerle:
@Dinonard 
> You only have one read :)

I guess reading constants are not a database read, so we don't count it?

Dinonard:
> @Dinonard
> 
> > You only have one read :)
> 
> I guess reading constants are not a database read, so we don't count it?

Right! There's no DB access here.

frame/dapps-staking/src/migrations.rs
Dinonard: You should remove this file from the PR.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -280,15 +285,15 @@ pub mod pallet {
...
             // Value is compared to 1 since genesis block is ignored
-            if now % blocks_per_era == BlockNumberFor::<T>::from(1u32)
-                || force_new_era
-                || previous_era.is_zero()
-            {
+            if now == next_era_starting_block || force_new_era || previous_era.is_zero() {
                 let next_era = previous_era + 1;
                 CurrentEra::<T>::put(next_era);
 
+                NextEraStartingBlock::<T>::put(now + blocks_per_era);
```
Dinonard, line 339: One of the parts of this task is to benchmark the on_initialize method (check the github issue ticket).

* https://docs.substrate.io/v3/runtime/benchmarking/
* check the benchmarking.rs file in pallet-dapps-staking module

jakehemmerle: I thought benchmarking is only for extrinsics, and not other functions like on_initialize(). I couldn't find any examples of non-extrinsics; can you link an example?

Dinonard:
@jakehemmerle I believe you're pushing to an incorrect branch.
Please push your commits here.

jakehemmerle:
> @jakehemmerle I believe you're pushing to an incorrect branch. Please push your commits here.

I was using an additional branch to handle benchmarking, which I would merge into here. I didn't want to mess up the state here trying to figure out how to do benchmarks.

jakehemmerle (commented):
I've bumped the pallet version and included some weights!

frame/dapps-staking/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-dapps-staking"
-version = "2.0.1"
```
Dinonard: You've added a new storage item so it's not a minor change anymore.
I think you can bump it up to 2.1.0.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -278,17 +283,17 @@ pub mod pallet {
...
-            let blocks_per_era = T::BlockPerEra::get();
             let previous_era = Self::current_era();
+            let next_era_starting_block = Self::next_era_starting_block();
 
             // Value is compared to 1 since genesis block is ignored
-            if now % blocks_per_era == BlockNumberFor::<T>::from(1u32)
-                || force_new_era
-                || previous_era.is_zero()
-            {
+            if now == next_era_starting_block || force_new_era || previous_era.is_zero() {
```
Dinonard: Small change of requirements - if we `pause` the pallet, it'd be good if normal operation was resumed as fast as possible once it's been unpaused. There's no need to wait for double or triple amount of blocks per era to pass before triggering a new one.

We can achieve this by changing the condition to `>=`.

Dinonard (commented):
Please run formatter

frame/dapps-staking/Cargo.toml
Dinonard: Yes :+1:

Dinonard (commented):
Great, well done!

Once the commit for runtimes is also ready for merge, let's merge them together.

jakehemmerle:
@Dinonard Cann you elaborate on what you mean by that? I believe this needs to be merged and tagged so I can update the dependencies in the runtime repo PR.

Dinonard:
Sure.

What I mean is that I'd like us to have the `Astar` repo code ready/approved before we merge this.
After you create a tag, you just update `Astar` commit and we merge it.

This is important since if you merge this, I need to rebase Individual claim on top of it which means OnRuntimeUpgrade
implementation needs to be available if I want to upgrade e.g. Shibuya.

jakehemmerle:
Note to self, make sure to update `pallet-precompile-dapps-staking` dep to point to new tagged `pallet-dapps-staking` after this PR gets merged.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -343,9 +348,11 @@ pub mod pallet {
 
                 Self::deposit_event(Event::<T>::NewDappStakingEra(next_era));
 
+                // should we return this or should we return a benchmarked weight?
                 consumed_weight + T::DbWeight::get().reads_writes(5, 3)
+                // return T::WeightInfo::on_initialize_new_era();
```
jakehemmerle: @Dinonard Do you want to just use the consumed_weight + read/writes or do you want to rebenchmark?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -343,9 +348,11 @@ pub mod pallet {
...
             } else {
-                T::DbWeight::get().reads(4)
+                return T::WeightInfo::on_initialize_not_new_era();
```
jakehemmerle: @Dinonard I don't think we need to rebenchmark this since we don't change any logic that happens if we don't have a new era.

frame/dapps-staking/src/pallet/mod.rs
jakehemmerle: @Dinonard

frame/dapps-staking/src/pallet/mod.rs
jakehemmerle: @Dinonard

Dinonard: This part is fine, nothing has been changed as you've noted.

But running benchmarks isn't really a big deal once you've built the binary.
Building the binary on the other hand takes time :P

frame/dapps-staking/src/pallet/mod.rs
Dinonard: This part is actually the reason I mentioned that you should have a few more days to figure this part out.
Because former benchmarks you've implemented are no longer valid for this scenario since we have the new `rotate_staking_info` part.

I will say that this depends on the approach you want to take.

First would be to redesign benchmarks so you can measure this correctly.

Another approach would be to just ditch the benchmarks and use it as I've added it.
(Btw, if you feel bad about loosing the benchmark code - no need. This was a learning exp).

jakehemmerle: Yes I remember you mentioning this! I'll take a look to conceptualize what the new benchmark might look like and will just keep your code here for simplicity.

frame/dapps-staking/src/pallet/mod.rs
jakehemmerle: https://xkcd.com/303/

frame/dapps-staking/src/pallet/mod.rs
Dinonard: So once you get a bit into this - let's consider how difficult it is (and how complex the code will be) to do proper benchmarking here.

It is also ok to just re-use what we already have (manual weight).

Also, keep in mind that it is possible to benchmark separate functions.
You don't need to benchmark an extrinsic call.
So it's perfectly fine to just do benchmarks on e.g. `rotate_staking_info`.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -345,7 +350,7 @@ pub mod pallet {
 
                 consumed_weight + T::DbWeight::get().reads_writes(5, 3)
             } else {
-                T::DbWeight::get().reads(4)
+                return T::WeightInfo::on_initialize_not_new_era();
```
Dinonard: Sorry if we misunderstood each other - just to repeat what we discussed on sync - we either go with fully-benchmarked solution or just go with _manual_ calculation.

Imo, you should drop the benchmarks :P

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -345,7 +350,7 @@ pub mod pallet {
 
                 consumed_weight + T::DbWeight::get().reads_writes(5, 3)
             } else {
-                T::DbWeight::get().reads(4)
+                T::DbWeight::get().reads(3)
```
Dinonard: You should keep it at 4:
* PalletDisabled::get
* Self::force_era
* Self::current_era
* Self::next_era_starting_block

frame/dapps-staking/src/tests.rs
```diff
@@ -1135,6 +1135,57 @@ fn withdraw_unbonded_no_unbonding_period() {
     })
 }
 
+#[test]
```
Dinonard: 2 requests for UT:
1. Please move them to the beginning of the file where `on_initialize` is tested
2. Add a UT that checks the _is greater_ part of `>=` check.

Dinonard (changes requested):
Please increment pallet version in the .toml file.

We can consider this as a new minor release.

frame/dapps-staking/src/pallet/mod.rs
jakehemmerle: We're on the same page! I don't think I had pushed the updates for this yet.

frame/dapps-staking/src/pallet/mod.rs
jakehemmerle: Ah, I missed force_era() !

frame/dapps-staking/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-dapps-staking"
-version = "3.0.4"
+version = "3.0.5"
```
Dinonard: This is a minor release bump so it should be `3.1.0`.

Please also sync the version of `pallet-precompiles-dapps-staking` with this one.
We decided to keep the versions the same since precompiles pallet has direct dependency towards `pallet-dapps-staking`.

frame/dapps-staking/src/tests.rs
```diff
@@ -91,6 +91,57 @@ fn on_initialize_is_ok() {
...
+        let start_block_number = System::block_number();
+
+        // go to start of next era
+        advance_to_era(mock::DappsStaking::current_era() + 1);
+
+        // show the length of the forced era is equal to blocks_per_era
+        let end_block_number = System::block_number();
+        assert_eq!(end_block_number - start_block_number, blocks_per_era);
+    })
+}
```
Dinonard, line 179: Thanks for moving the UTs!

Please also sort out the other part of the comment.

The best way to do this is:
* prepare a new UT
* initialize first block
* enable maintenance mode
* advance to era (current_era + era_length + 3)  // or some other number
* verify that current block is higher than `NextEraStartingBlock` but that era hasn't changed
* disable maintenance mode
* advance one era
* verify era has changed and `NextEraStartingBlock` has expected value

Dinonard (changes requested):
Sort out the two remaining comments and let's merge this.

frame/dapps-staking/src/tests.rs
jakehemmerle: Ah yes, I missed the second part of the comment. Will address this today!

frame/dapps-staking/src/tests.rs
```diff
@@ -91,6 +91,93 @@ fn on_initialize_is_ok() {
...
+
+#[test]
+fn new_era_is_handled_with_maintenance_mode() {
+    ExternalityBuilder::build().execute_with(|| {
+        initialize_first_block();
+
+        // enable maintenance mode
+        assert_ok!(DappsStaking::ensure_pallet_enabled());
+        assert!(!PalletDisabled::<TestRuntime>::exists());
+
```
Dinonard: Not needed for this test

Dinonard:
Fix formatting and it can be merged.

Dinonard (approved):
LGTM

PR #6, opened 2022-02-03
Individual Claim

Dinonard:
**Pull Request Summary**

Main purpose of this commit is to solve scalability problems with the former `claim` implementation.
Instead of using group claims, each staker will now be responsible for claiming their own rewards.

This entails changes in storage schema and other extrinsic calls.
* dApp state is now tracked using a dedicated enum `DAppState`
* Cleaned `Forcing` enum of unused values
* `EraStakingPoints` now contains info about staked value, number of stakers and whether dapp rewards were claimed for an era
* `StakerInfo` is a new struct which handles `(staker, contract)` info storage in a scalable manner
  * the idea is to wrap history into a vector of limited size which prevents stakers from overflowing our DB
  * methods are available for easier handling, extensive and thorough UTs provided
 * Introduced `RewardHandling` enum as preparation for future task (to avoid storage migration)
 * `HistoryDepth` limit has been removed
 * `BonusEraDuration` has been removed as it serves no purpose anymore
 * New extrinsic `withdraw_from_unregistered` to cleanup stakes after contract was removed
 * claiming functionality now has separate calls for `claim_staker` and `claim_dapp`
 * very extensive UTs provided

**TODO**
- [x] rename EraStakeAndReward to something more appropriate
- [x] manual test on local node
- [x] update precompiles code
- [x] update documentation (README)
- [x] re-run all benchmarks prior to merge
- [x] implement new benchmarking
- [x] implement & test storage migration
- [x] cover code with UTs
- [x] implement individual claim 

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


frame/dapps-staking/src/lib.rs
```diff
@@ -97,6 +130,128 @@ impl Default for Version {
...
+
+    /// Stakes some value in the specified era.
+    /// User should ensure that given era is either equal or greater than the
+    /// latest available era in the staking info.
+    fn stake(&mut self, current_era: EraIndex, value: Balance) -> Result<(), &str> {
+        if self.stakes.is_empty() {
+            self.stakes.push(EraStake::new(value, current_era))
+        } else {
+            let era_stake = self.stakes.last().unwrap(); // exists if vec not empty
+            if era_stake.era > current_era {
```
Maar-io, line 220: When can this happen?

Dinonard: Initially, I implemented `stake` and `unstake` without this check.
But since we have to be 100% sure nothing can get messed up, I added this check.

To answer your question - it cannot happen :).
It's just a safety measure in case I have a bug in the code that would allow it.

frame/dapps-staking/src/lib.rs
```diff
@@ -97,6 +130,128 @@ impl Default for Version {
...
+        }
+
+        let new_stake_value = era_stake.staked.saturating_sub(value);
+
+        if current_era == era_stake.era {
+            *self.stakes.last_mut().unwrap() = EraStake::new(new_stake_value, current_era)
+        } else {
+            self.stakes
+                .push(EraStake::new(new_stake_value, current_era))
+        }
```
Maar-io: code duplication? same is in stake()

frame/dapps-staking/src/benchmarking.rs
```diff
@@ -180,22 +195,37 @@ benchmarks! {
...
+        assert!(era > claim_era);
+    }
 
+    claim_dapp {
         initialize::<T>();
-        let (developer_id, contract_id) = register_contract::<T>()?;
+        let (developer, contract_id) = register_contract::<T>()?;
 
-        let number_of_stakers = n - 1;
+        let number_of_stakers = 5;
```
hoonsubin: Can you tell me why the number of stakers are set as a constant?

Dinonard: Leftover from previous implementation when weight was dependent on the number of active stakers.
I could change it to e.g. 1 or just remove this (immutable) variable.

frame/dapps-staking/src/lib.rs
```diff
@@ -97,6 +130,128 @@ impl Default for Version {
...
+    /// `Claims` the oldest era available for claiming.
+    /// In case valid era exists, returns (claim era, staked amount) tuple.
+    /// If no valid era exists, returns (0, 0) tuple.
+    fn claim(&mut self) -> (EraIndex, Balance) {
+        if self.is_empty() {
+            (0, Zero::zero())
+        } else {
+            let era_stake = self.stakes[0]; // not empty so it exists
+
+            if self.stakes.len() == 1 || self.stakes[1].era > era_stake.era + 1 {
```
Maar-io, line 310: What if `self.stakes[1]` does not exist?

frame/dapps-staking/src/lib.rs
Dinonard: Good catch!
I tried to avoid this by wrapping it up in a method but the problem is I cannot make it private.
And I don't want to expose this as part of the API.

E.g. it would be something like 'update_staked_value(era, value)`

frame/dapps-staking/src/lib.rs
```diff
@@ -196,12 +351,39 @@ where
...
+    /// Rewards are transfered to stakers balance without any further action.
+    OnlyPayout,
+    /// Rewards are transfered to stakers balance and are immediately re-staked
+    /// on the contract from which the reward was received.
+    PayoutAndStake,
+}
+
+impl Default for RewardHandling {
+    fn default() -> Self {
+        RewardHandling::PayoutAndStake
```
Maar-io: Shouldn't the default be `OnlyPayout`?

frame/dapps-staking/src/lib.rs
Dinonard: That code will never execute if the first part of the clause is true.

If `self.is_empty()` is false and `self.stakes.len() == 1` is false, it means that `self.stakes.len()` is greater than 1.
So `self.stakes[1]' will always exist if that part of the code is executed.

frame/dapps-staking/src/lib.rs
Dinonard: We can discuss this.

My motivation is that we should make `PayoutAndStake` default because this is what benefits users more, as well as the ecosystem.

This can be changed in the commit that will add actual support for this.

frame/dapps-staking/src/lib.rs
```diff
@@ -24,6 +27,8 @@ mod mock;
 mod testing_utils;
 #[cfg(test)]
 mod tests;
+#[cfg(test)]
+mod tests_lib;
 
 pub use pallet::pallet::*;
 pub use sp_staking::SessionIndex;
```
Maar-io: Unused. you can delete

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -371,14 +387,26 @@ pub mod pallet {
         /// This must be called by the developer who registered the contract.
         ///
         /// Warning: After this action contract can not be assigned again.
-        #[pallet::weight(T::WeightInfo::unregister(T::MaxNumberOfStakersPerContract::get()))]
+        #[pallet::weight(T::WeightInfo::unregister())]
         pub fn unregister(
```
Maar-io, line 423: once we get rid of Sudo, how can an admin do unregister for contracts that prove to not be worked on or malicious?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -371,14 +387,26 @@ pub mod pallet {
...
+            ensure!(
+                dapp_info.state == DAppState::Registered,
+                Error::<T>::NotOperatedContract
+            );
+            ensure!(
+                dapp_info.developer == developer,
+                Error::<T>::NotOwnedContract,
+            );
+
+            // TODO: Is this really needed? Why shouldn't same account be eligible to register multiple contracts?
```
Maar-io: perhaps spam protection?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -389,39 +417,68 @@ pub mod pallet {
...
 
-            let number_of_stakers = staking_info.stakers.len();
-            Ok(Some(T::WeightInfo::unregister(number_of_stakers as u32)).into())
+            Ok(().into())
+        }
+
+        /// Withdraw locked funds from a contract that was unregistered.
+        /// Funds don't need to undergo the unbonding period - they are returned immediately.
+        #[pallet::weight(T::WeightInfo::withdraw_from_unregistered())]
+        pub fn withdraw_from_unregistered(
```
Maar-io, line 452: @bobo-k2 how do you plan to implement this on frontend?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -625,105 +681,136 @@ pub mod pallet {
...
         }
 
-        /// Claim the rewards earned by contract_id.
-        /// All stakers and developer for this contract will be paid out with single call.
-        /// claim is valid for all unclaimed eras but not longer than history_depth().
-        /// Any reward older than history_depth() will go to Treasury.
-        /// Any user can call this function.
-        #[pallet::weight(T::WeightInfo::claim(T::MaxNumberOfStakersPerContract::get() + 1))]
-        pub fn claim(
+        // TODO: do we need to add force methods or at least methods that allow others to claim for someone else?
```
Maar-io, line 720: once we implement re-stake, projects might be interested to claim in name of the stakers and thus increase the staked amount for their contracts

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -625,105 +681,136 @@ pub mod pallet {
...
+            origin: OriginFor<T>,
+            contract_id: T::SmartContract,
+        ) -> DispatchResultWithPostInfo {
+            ensure!(!Self::pallet_disabled(), Error::<T>::Disabled);
+            let staker = ensure_signed(origin)?;
+
+            // Ensure we have something to claim
+            let mut staker_info = Self::staker_info(&staker, &contract_id);
+            let (era, staked) = staker_info.claim();
+            ensure!(staked > Zero::zero(), Error::<T>::NotStakedContract);
```
Maar-io, line 734: `staked != Zero::zero()` maybe?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -625,105 +681,136 @@ pub mod pallet {
...
+
+            let current_era = Self::current_era();
+
+            ensure!(era < current_era, Error::<T>::EraOutOfBounds);
+
+            let staking_info = Self::staking_info(&contract_id, era);
+            let reward_and_stake =
+                Self::era_reward_and_stake(era).ok_or(Error::<T>::UnknownEraReward)?;
+
+            let (_, stakers_reward) = Self::dev_stakers_split(
```
Maar-io: rename to `stakers_joint_reward `

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -625,105 +681,136 @@ pub mod pallet {
...
+                era,
+                staker_reward,
+            ));
+
+            Ok(().into())
+        }
+
+        /// Claim earned dapp rewards for the specified era.
+        #[pallet::weight(T::WeightInfo::claim_dapp())]
+        pub fn claim_dapp(
```
Maar-io, line 777: I'd still stick for a single claim call (stakers + dev) for the front-end but it is up to you and @bobo-k2

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Sure, I will change it to root-only.
We will also need a couple for root-only force methods in addition to this.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: We have two levels of spam protection:
- dev needs to be whitelisted
- each registration requires reservation amount

It'd be a very expensive spam.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Sure, I was thinking the same thing.

It's easy to add a wrapped call, no problem there.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: I'd prefer `!staked.is_zero()` but the when we got audited by Quantstamp, they explicitly told me to prefer **is greater than** check.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: For the frontend, I agree. Not part of this commit though.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: I can implement wrappers after I get review from Aleks.

frame/dapps-staking/src/lib.rs
```diff
@@ -193,39 +229,77 @@ impl<Balance: AtLeast32BitUnsigned + Copy> StakerInfo<Balance> {
...
-        }
+        if let Some(era_stake) = self.stakes.last() {
+            if era_stake.era > current_era {
+                return Err("Unexpected era".into());
+            }
 
-        let new_stake_value = era_stake.staked.saturating_sub(value);
+            let new_stake_value = era_stake.staked.saturating_sub(value);
+            if current_era == era_stake.era {
+                *self.stakes.last_mut().unwrap() = EraStake::new(new_stake_value, current_era)
```
Dinonard: Improve this too

frame/dapps-staking/src/lib.rs
```diff
@@ -101,11 +130,239 @@ impl<AccountId: Ord, Balance: HasCompact + Default> Default
...
+    ///    Internal vector is modified to `[<15, 3000>]`
+    ///    Note that `0` staked period is discarded since nothing can be claimed there.
+    ///
+    /// 4. `claim()` will return `(15, 3000)`.
+    ///    Internal vector is modified to `[16, 3000]`
+    ///
+    /// Repeated calls would continue to modify vector following the same rule as in *4.*
+    ///
+    fn claim(&mut self) -> (EraIndex, Balance) {
+        if self.is_empty() {
```
akru: Let's use `if let` here too.

frame/dapps-staking/src/lib.rs
akru: Could you explain this condition, in that case `self.stakes[1].era > era_stake.era + 1` is not true?

frame/dapps-staking/src/lib.rs
```diff
@@ -101,11 +130,239 @@ impl<AccountId: Ord, Balance: HasCompact + Default> Default
...
+    /// Latest staked value.
+    /// E.g. if staker is fully unstaked, this will return `Zero`.
+    /// Othwerise returns a non-zero balance.
+    pub fn latest_staked_value(&self) -> Balance {
+        self.stakes.last().map_or(Zero::zero(), |x| x.staked)
+    }
+
+    /// Removes unstaked values if they're no longer valid for comprehension
+    fn clean_unstaked(&mut self) {
+        if !self.stakes.is_empty() && self.stakes[0].staked.is_zero() {
```
akru: May be we should include this logic into method that use it, it's not too heavy to use dedicated method but readability will increase if we move this operations into methods instead of alias.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -178,7 +196,19 @@ pub mod pallet {
         T::SmartContract,
         Twox64Concat,
         EraIndex,
-        EraStakingPoints<T::AccountId, BalanceOf<T>>,
+        ContractStakeInfo<BalanceOf<T>>,
+    >;
+
+    #[pallet::storage]
+    #[pallet::getter(fn staker_info)]
+    pub(crate) type StakersInfo<T: Config> = StorageDoubleMap<
```
akru: May be rename `StakersInfo` -> `GeneralStakerInfo`, because of possible types `Stakers` = `Staker`

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -372,57 +416,87 @@ pub mod pallet {
...
         pub fn unregister(
             origin: OriginFor<T>,
             contract_id: T::SmartContract,
         ) -> DispatchResultWithPostInfo {
-            ensure!(!Self::pallet_disabled(), Error::<T>::Disabled);
-            let developer = ensure_signed(origin)?;
-
-            let registered_contract =
-                RegisteredDevelopers::<T>::get(&developer).ok_or(Error::<T>::NotOwnedContract)?;
+            Self::ensure_pallet_enabled()?;
```
akru, line 427: I'm not sure that root calls should have this guard

frame/dapps-staking/src/pallet/mod.rs
akru: @Dinonard could you explain please why it should be dedicated method? It looks like a case for general unbond method.

frame/dapps-staking/src/lib.rs
Dinonard: Sure, no problem.

frame/dapps-staking/src/lib.rs
Dinonard: I did before but didn't like how it looked...but I'll push the change again.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: This is intentional - I will use maintenance mode to prevent any storage access during storage migration.
`unregister` is also just a normal extrinsic.

If there's a need to use `unregister` during maintenance mode (imo there shouldn't be), we can just wrap it into a batch call where we `disable maintenance - execute unregister - enable maintenance`

frame/dapps-staking/src/pallet/mod.rs
Dinonard: It's actually quite different - the checks differ.

Here we require that all pending rewards have been claimed. Also, there's no unbonding period here.

Sure, I could put it all into `unbond_and_unstake` but it creates a more complicated method with many **if-s**.
This, in my opinion, is much better. Code is easier to test and there is less room for error.

akru: Clean interface is also important. Not sure why withdraw from unregistered dapps should be different from general withdraw.

frame/dapps-staking/README.md
```diff
@@ -3,13 +3,13 @@ This document describes the interface for the pallet-dapps-staking.
...
 Table of Contents:
 1. [Terminology](#Terminology)
-1. [Types](#Types)
-1. [Events](#Events)
-1. [Errors](#Errors)
-1. [Calls](#Calls)
-1. [Storage](#Storage)
-1. [Referent implementatio](#Referent)
-1. [FAQ](#FAQ)
+2. [Types](#Types)
```
codespool: Why keep the table of contents if there's no content anymore?

Dinonard: Good catch, I will remove it.

frame/dapps-staking/src/lib.rs
Dinonard: We have two possible branches:
* first one is when we only have a single element in the vec OR if the next era for which we have info isn't the consecutive era which we are claiming. E.g.:
<1, 1000>, <5, 2000> ==> <2, 1000>, <5, 2000>
*In this scenario, we just modify the **first** element by changing its era
* the second scenario can only be the case where the next element in vec is the consecutive era
* this means we can just delete the first element since that information isn't relevant anymore
<1, 1000>, <2, 2000> ==> <2, 2000>

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Sure, I can do that.

@impelcrypto please not that I will change this in the new version.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: I agree with you in that statement but I don't want to make code more complicated - by doing this split, there is no branching, most if/else clauses are avoided.

And the behavior is inherently different.

If we use `unbond_and_unstake` we can `withdaraw_unbonded` after unbonding period has passed.
However, `withdraw_from_unregistered` acts as `unbond_unstake_and_withdraw` all at once.
So the behavior is quite different.

There is no unbonding period here - funds are returned directly to the user.
Also, you cannot call this if there are pending rewards - this is one of the simplifications I made to avoid complexity in the `StakerInfo`.

Should we maybe rename it?

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Btw, perhaps looking at the test utility method for `withdraw_from_unregistered` would help. It think it shows how behavior is different.

codespool:
Found this while fixing my own PR:
you're still using reward_handling property name in migrations:
https://github.com/AstarNetwork/astar-frame/blob/6df33856f0616fcecf21e5bb593a444961c9bbe5/frame/dapps-staking/src/migrations.rs#L548
is that intentional?

Dinonard:
> Found this while fixing my own PR: you're still using reward_handling property name in migrations:
> 
> https://github.com/AstarNetwork/astar-frame/blob/6df33856f0616fcecf21e5bb593a444961c9bbe5/frame/dapps-staking/src/migrations.rs#L548
> 
> is that intentional?

Oh, I didn't rename the variable name. I will do it now.
The type is `RewardDestination` though, even here.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -775,11 +873,31 @@ pub mod pallet {
...
             ensure_root(origin)?;
             PreApprovalIsEnabled::<T>::put(enabled);
             Ok(().into())
         }
+
+        /// `true` will disable pallet, enabling maintenance mode. `false` will do the opposite.
+        #[pallet::weight(T::WeightInfo::maintenance_mode())]
+        pub fn maintenance_mode(
+            origin: OriginFor<T>,
+            enable_maintenace: bool,
```
codespool: typo :)

Dinonard: fixed

Maar-io (approved):
LGTM

Dinonard:
@codespool You can rebase on the default branch now. No more force-pushes :D

PR #7, opened 2022-02-07
Staking precompile removal

Dinonard:
**Pull Request Summary**

Removed  staking precompile from the workspace.

**Check list**
- [ ] adds new feature
- [ ] updated spec version
- [ ] relies on other tasks
- [ ] documentation changes

**Fixes**
- compilation issues by removing the culprit code from the workspace
- if agreed, we can remove the pallet code entirely later

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macOS-latest]
+        os: [ubuntu-latest]
     steps:
     - uses: hecrj/setup-rust-action@v1
       with:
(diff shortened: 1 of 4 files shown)


.github/workflows/integration.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macOS-latest]
+        os: [ubuntu-latest]
```
Maar-io, line 13: why you removed mac?

Maar-io (approved):

.github/workflows/integration.yml
Dinonard: Because we're not doing a build, only a check.
And mac is more _expensive_ and slow so I thought this would save us some time.

astar-collator will still be built for mac, that remains unchanged.

PR #9, opened 2022-02-10
Added Sr25519Precompile 

akru:
**Pull Request Summary**

This precompile helps to verify Sr25519 signatures in EVM smart contracts.

Proposed interface:

```solidity
interface SR25519 {
    /**
     * @dev Verify signed message using SR25519 crypto.
     * @return A boolean confirming whether the public key is signer for the message. 
     */
    function verify(
        bytes32 public,
        bytes signature,
        bytes message
    ) external view returns (bool);
}
```

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)


precompiles/utils/macro/src/lib.rs
```diff
@@ -0,0 +1,144 @@
+// Copyright 2019-2022 PureStake Inc.
+// This file is part of Moonbeam.
+
+// Moonbeam is free software: you can redistribute it and/or modify
```
hoonsubin, line 4: You might want to change the headers...

akru: Because of use it as is I guess header should also keep for legal reasons.

Dinonard: In that case, would it be better to fork their repo and just add our modifications there?

Not that keen on having another `frame` repo to depend on but this header in such form also doesn't look appealing.

Could we just change it to something like "re-used from Moonbeam repository <link>"?

Maar-io:
please change PR to point to branch polkadot-v0.9.16

akru:
@Maar-io changed and resolved conflicts

Maar-io:
@akru test was failing, I added the fix

Maar-io (approved):
LGTM

PR #14, opened 2022-02-16
Feature/compound reward

codespool:
**Pull Request Summary**

> Adds automatic restaking of reward to `claim_staker` extrinsic by checking the `ledger.reward_handling` property and staking the claimed amount right after claim.

**Check list**
- [X] adds new feature
- [X] modifies existing feature (bug fix or improvements)
- [ ] relies on other tasks
- [ ] documentation changes
- [X] tests and/or benchmarks are included

**Adds**
- `ledger.reward_handling` check to `claim_staker`
- `RewardAndRestake`, `RewardHandlingChange` events
- `enable_compound_staking` extrinsic
- `update_staking_values` helper function
- tests to check claiming after unstake or manually switching `reward_handling`

**Fixes**
- fix `assert_claim_staker` test utility function to take into account `reward_handling` state 

**Changes**
- `claim_staker` now checks for `reward_handling` state, and restakes the claimed value if needed
- `bond_and_stake` uses `update_staking_values` helper function

**To-dos**
- [ ] update docs

### DIFF
--- a/frame/dapps-staking/src/benchmarking.rs
+++ b/frame/dapps-staking/src/benchmarking.rs
@@ -243,7 +243,8 @@ benchmarks! {
     }: _(RawOrigin::Root, true)
     enable_compound_staking {
         initialize::<T>();
-        let staker = whitelisted_caller();
+
+        let staker: T::AccountId = whitelisted_caller();
         let option = RewardHandling::OnlyPayout;
     }: _(RawOrigin::Signed(staker.clone()), option)
     verify {
--- a/frame/dapps-staking/src/benchmarking.rs
+++ b/frame/dapps-staking/src/benchmarking.rs
@@ -212,6 +212,8 @@ benchmarks! {
         let claim_era = DappsStaking::<T>::current_era();
         let stakers = prepare_bond_and_stake::<T>(number_of_stakers, &contract_id, SEED)?;
         let staker = stakers[0].clone();
+
+        DappsStaking::<T>::set_reward_destination(RawOrigin::Signed(staker.clone()).into(), RewardDestination::StakeBalance)?;
         advance_to_era::<T>(claim_era + 1u32);
 
     }: claim_staker(RawOrigin::Signed(staker.clone()), contract_id.clone())
--- a/frame/dapps-staking/src/benchmarking.rs
+++ b/frame/dapps-staking/src/benchmarking.rs
@@ -268,7 +268,7 @@ benchmarks! {
         let option = RewardDestination::FreeBalance;
     }: _(RawOrigin::Signed(staker.clone()), option)
     verify {
-        assert_last_event::<T>(Event::<T>::RewardDestinationSet(staker, option).into());
+        assert_last_event::<T>(Event::<T>::RewardDestination(staker, option).into());
     }
 
 }
--- a/frame/dapps-staking/src/benchmarking.rs
+++ b/frame/dapps-staking/src/benchmarking.rs
@@ -204,7 +204,7 @@ benchmarks! {
         assert_last_event::<T>(Event::<T>::Withdrawn(staker, unstake_amount).into());
     }
 
-    claim_staker {
+    claim_staker_with_restake {
         initialize::<T>();
         let (_, contract_id) = register_contract::<T>()?;
 
@@ -214,7 +214,26 @@ benchmarks! {
         let staker = stakers[0].clone();
         advance_to_era::<T>(claim_era + 1u32);
 
-    }: _(RawOrigin::Signed(staker.clone()), contract_id.clone())
+    }: claim_staker(RawOrigin::Signed(staker.clone()), contract_id.clone())
+    verify {
+        let mut staker_info = DappsStaking::<T>::staker_info(&staker, &contract_id);
+        let (era, _) = staker_info.claim();
+        assert!(era > claim_era);
+    }
+
+    claim_staker_without_restake {
+        initialize::<T>();
+        let (_, contract_id) = register_contract::<T>()?;
+
+        let number_of_stakers = 3;
+        let claim_era = DappsStaking::<T>::current_era();
+        let stakers = prepare_bond_and_stake::<T>(number_of_stakers, &contract_id, SEED)?;
+        let staker = stakers[0].clone();
+
+        DappsStaking::<T>::set_reward_destination(RawOrigin::Signed(staker.clone()).into(), RewardDestination::FreeBalance)?;
+        advance_to_era::<T>(claim_era + 1u32);
+
+    }: claim_staker(RawOrigin::Signed(staker.clone()), contract_id.clone())
     verify {
         let mut staker_info = DappsStaking::<T>::staker_info(&staker, &contract_id);
         let (era, _) = staker_info.claim();
@@ -249,7 +268,7 @@ benchmarks! {
         let option = RewardDestination::FreeBalance;
     }: _(RawOrigin::Signed(staker.clone()), option)
     verify {
-        assert_last_event::<T>(Event::<T>::RewardChange(staker, option).into());
+        assert_last_event::<T>(Event::<T>::RewardDestinationSet(staker, option).into());
     }
 
 }
--- a/frame/dapps-staking/src/benchmarking.rs
+++ b/frame/dapps-staking/src/benchmarking.rs
@@ -241,6 +241,14 @@ benchmarks! {
 
     maintenance_mode {
     }: _(RawOrigin::Root, true)
+    enable_compound_staking {
+        initialize::<T>();
+        let staker = whitelisted_caller();
+        let option = RewardHandling::OnlyPayout;
+    }: _(RawOrigin::Signed(staker.clone()), option)
+    verify {
+        assert_last_event::<T>(Event::<T>::RewardHandlingChange(staker, option).into());
+    }
 
 }
 
--- a/frame/dapps-staking/src/benchmarking.rs
+++ b/frame/dapps-staking/src/benchmarking.rs
@@ -241,14 +241,15 @@ benchmarks! {
 
     maintenance_mode {
     }: _(RawOrigin::Root, true)
-    enable_compound_staking {
+
+    set_reward_destination {
         initialize::<T>();
 
         let staker: T::AccountId = whitelisted_caller();
-        let option = RewardHandling::OnlyPayout;
+        let option = RewardDestination::FreeBalance;
     }: _(RawOrigin::Signed(staker.clone()), option)
     verify {
-        assert_last_event::<T>(Event::<T>::RewardHandlingChange(staker, option).into());
+        assert_last_event::<T>(Event::<T>::RewardChange(staker, option).into());
     }
 
 }
--- a/frame/dapps-staking/src/benchmarking.rs
+++ b/frame/dapps-staking/src/benchmarking.rs
@@ -266,8 +266,12 @@ benchmarks! {
     set_reward_destination {
         initialize::<T>();
 
-        let staker: T::AccountId = whitelisted_caller();
         let option = RewardDestination::FreeBalance;
+        let (_, contract_id) = register_contract::<T>()?;
+
+        let number_of_stakers = 1;
+        let stakers = prepare_bond_and_stake::<T>(number_of_stakers, &contract_id, SEED)?;
+        let staker = stakers[0].clone();
     }: _(RawOrigin::Signed(staker.clone()), option)
     verify {
         assert_last_event::<T>(Event::<T>::RewardDestination(staker, option).into());
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -511,12 +511,16 @@ pub mod pallet {
             );
 
             // Increment ledger and total staker value for contract.
-            Self::update_staking_values(
-                value_to_stake,
-                &mut ledger,
-                &mut staking_info,
-                current_era,
-            )?;
+            ledger.locked = ledger.locked.saturating_add(value_to_stake);
+            staking_info.total = staking_info.total.saturating_add(value_to_stake);
+
+            // Update storage
+            GeneralEraInfo::<T>::mutate(&current_era, |value| {
+                if let Some(x) = value {
+                    x.staked = x.staked.saturating_add(value_to_stake);
+                    x.locked = x.locked.saturating_add(value_to_stake);
+                }
+            });
 
             Self::update_ledger(&staker, ledger);
             Self::update_staker_info(&staker, &contract_id, staker_info);
@@ -708,12 +712,15 @@ pub mod pallet {
                     .stake(current_era, staker_reward)
                     .map_err(|_| Error::<T>::UnexpectedStakeInfoEra)?;
 
-                Self::update_staking_values(
-                    staker_reward,
-                    &mut ledger,
-                    &mut staking_info,
-                    current_era,
-                )?;
+                ledger.locked = ledger.locked.saturating_add(staker_reward);
+                staking_info.total = staking_info.total.saturating_add(staker_reward);
+                // Update storage
+                GeneralEraInfo::<T>::mutate(&current_era, |value| {
+                    if let Some(x) = value {
+                        x.staked = x.staked.saturating_add(staker_reward);
+                        x.locked = x.locked.saturating_add(staker_reward);
+                    }
+                });
 
                 Self::update_ledger(&staker, ledger);
 
@@ -896,25 +903,6 @@ pub mod pallet {
             } else {
                 Ok(())
             }
-        }
-
-        fn update_staking_values(
-            value_to_stake: BalanceOf<T>,
-            ledger: &mut AccountLedger<BalanceOf<T>>,
-            staking_info: &mut EraStakingPoints<BalanceOf<T>>,
-            current_era: u32,
-        ) -> DispatchResultWithPostInfo {
-            ledger.locked = ledger.locked.saturating_add(value_to_stake);
-            staking_info.total = staking_info.total.saturating_add(value_to_stake);
-            // Update storage
-            GeneralEraInfo::<T>::mutate(&current_era, |value| {
-                if let Some(x) = value {
-                    x.staked = x.staked.saturating_add(value_to_stake);
-                    x.locked = x.locked.saturating_add(value_to_stake);
-                }
-            });
-
-            Ok(().into())
         }
 
         /// Update the ledger for a staker. This will also update the stash lock.
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -130,6 +130,11 @@ pub mod pallet {
     #[pallet::getter(fn force_era)]
     pub type ForceEra<T> = StorageValue<_, Forcing, ValueQuery, ForceEraOnEmpty>;
 
+    /// Stores the block number of when the next era starts
+    #[pallet::storage]
+    #[pallet::getter(fn next_era_starting_block)]
+    pub type NextEraStartingBlock<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
+
     /// Registered developer accounts points to corresponding contract
     #[pallet::storage]
     #[pallet::getter(fn registered_contract)]
@@ -141,12 +146,6 @@ pub mod pallet {
     #[pallet::getter(fn dapp_info)]
     pub(crate) type RegisteredDapps<T: Config> =
         StorageMap<_, Blake2_128Concat, T::SmartContract, DAppInfo<T::AccountId>>;
-
-    /// Legacy, don't use.
-    /// TODO: remove in future upgrades
-    #[pallet::storage]
-    pub type EraRewardsAndStakes<T: Config> =
-        StorageMap<_, Twox64Concat, EraIndex, migrations::v3::OldEraRewardAndStake<BalanceOf<T>>>;
 
     /// Total staked, locked & rewarded for a particular era
     #[pallet::storage]
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -493,9 +493,6 @@ pub mod pallet {
                 staking_info.number_of_stakers = staking_info.number_of_stakers.saturating_add(1);
             }
 
-            staker_info
-                .stake(current_era, value_to_stake)
-                .map_err(|_| Error::<T>::UnexpectedStakeInfoEra)?;
             ensure!(
                 // One spot should remain for compounding reward claim call
                 staker_info.len() < T::MaxEraStakeValues::get(),
@@ -507,26 +504,15 @@ pub mod pallet {
             );
 
             // Increment ledger and total staker value for contract. Overflow shouldn't be possible but the check is here just for safety.
-            ledger.locked = ledger
-                .locked
-                .checked_add(&value_to_stake)
-                .ok_or(ArithmeticError::Overflow)?;
-            staking_info.total = staking_info
-                .total
-                .checked_add(&value_to_stake)
-                .ok_or(ArithmeticError::Overflow)?;
-
-            // Update storage
-            GeneralEraInfo::<T>::mutate(&current_era, |value| {
-                if let Some(x) = value {
-                    x.staked = x.staked.saturating_add(value_to_stake);
-                    x.locked = x.locked.saturating_add(value_to_stake);
-                }
-            });
-
-            Self::update_ledger(&staker, ledger);
-            Self::update_staker_info(&staker, &contract_id, staker_info);
-            ContractEraStake::<T>::insert(&contract_id, current_era, staking_info);
+            let stake_result = Self::stake(
+                &staker,
+                &contract_id,
+                &mut ledger,
+                value_to_stake,
+                &mut staking_info,
+                &mut staker_info,
+            );
+            ensure!(stake_result.is_ok(), stake_result.err());
 
             Self::deposit_event(Event::<T>::BondAndStake(
                 staker,
@@ -713,6 +699,18 @@ pub mod pallet {
                 staker_reward,
             ));
 
+            let mut ledger = Self::ledger(&staker);
+            if let RewardHandling::PayoutAndStake = ledger.reward_handling {
+                let stake_result = Self::stake(
+                    &staker,
+                    &contract_id,
+                    &mut ledger,
+                    staker_reward,
+                    &mut staking_info,
+                    &mut staker_info,
+                );
+                ensure!(stake_result.is_ok(), stake_result.err());
+            }
             Ok(().into())
         }
 
@@ -865,6 +863,42 @@ pub mod pallet {
             }
         }
 
+        fn stake(
+            staker: &<T as frame_system::Config>::AccountId,
+            contract_id: &T::SmartContract,
+            ledger: &mut AccountLedger<BalanceOf<T>>,
+            value_to_stake: BalanceOf<T>,
+            staking_info: &mut EraStakingPoints<BalanceOf<T>>,
+            staker_info: &mut StakerInfo<BalanceOf<T>>,
+        ) -> DispatchResultWithPostInfo {
+            let current_era = Self::current_era();
+            staker_info
+                .stake(current_era, value_to_stake)
+                .map_err(|_| Error::<T>::UnexpectedStakeInfoEra)?;
+
+            ledger.locked = ledger
+                .locked
+                .checked_add(&value_to_stake)
+                .ok_or(ArithmeticError::Overflow)?;
+            staking_info.total = staking_info
+                .total
+                .checked_add(&value_to_stake)
+                .ok_or(ArithmeticError::Overflow)?;
+
+            // Update storage
+            GeneralEraInfo::<T>::mutate(&current_era, |value| {
+                if let Some(x) = value {
+                    x.staked = x.staked.saturating_add(value_to_stake);
+                    x.locked = x.locked.saturating_add(value_to_stake);
+                }
+            });
+
+            Self::update_ledger(&staker, ledger.clone());
+            Self::update_staker_info(&staker, &contract_id, staker_info.clone());
+            ContractEraStake::<T>::insert(contract_id.clone(), current_era, staking_info);
+
+            Ok(().into())
+        }
         /// Update the ledger for a staker. This will also update the stash lock.
         /// This lock will lock the entire funds except paying for further transactions.
         fn update_ledger(staker: &T::AccountId, ledger: AccountLedger<BalanceOf<T>>) {
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -275,6 +275,10 @@ pub mod pallet {
         RequiredContractPreApproval,
         /// Developer's account is already part of pre-approved list
         AlreadyPreApprovedDeveloper,
+        /// No ledger info for account
+        NothingStakedForAccount,
+        /// Reward handling already set to the requested value
+        RestakingAlreadySet,
     }
 
     #[pallet::hooks]
@@ -863,6 +867,30 @@ pub mod pallet {
 
             Self::deposit_event(Event::<T>::MaintenanceMode(enable_maintenance));
 
+        #[pallet::weight(T::WeightInfo::enable_compound_staking())]
+        pub fn enable_compound_staking(
+            origin: OriginFor<T>,
+            enabled: bool,
+        ) -> DispatchResultWithPostInfo {
+            ensure!(!Self::pallet_disabled(), Error::<T>::Disabled);
+            let staker = ensure_signed(origin)?;
+            let mut ledger = Self::ledger(&staker);
+
+            ensure!(!ledger.is_empty(), Error::<T>::NothingStakedForAccount);
+
+            let requested_reward_handling = if enabled {
+                RewardHandling::PayoutAndStake
+            } else {
+                RewardHandling::OnlyPayout
+            };
+            ensure!(
+                requested_reward_handling != ledger.reward_handling,
+                Error::<T>::RestakingAlreadySet
+            );
+
+            ledger.reward_handling = requested_reward_handling;
+
+            Self::update_ledger(&staker, ledger);
             Ok(().into())
         }
     }
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -221,7 +221,7 @@ pub mod pallet {
         /// Maintenance mode has been enabled or disabled
         MaintenanceMode(bool),
         /// Claimed staker reward restaked
-        RewardRestaked(T::AccountId, T::SmartContract, BalanceOf<T>),
+        RewardAndRestake(T::AccountId, T::SmartContract, BalanceOf<T>),
     }
 
     #[pallet::error]
@@ -276,10 +276,6 @@ pub mod pallet {
         RequiredContractPreApproval,
         /// Developer's account is already part of pre-approved list
         AlreadyPreApprovedDeveloper,
-        /// No ledger info for account
-        NothingStakedForAccount,
-        /// Reward handling already set to the requested value
-        RestakingAlreadySet,
     }
 
     #[pallet::hooks]
@@ -721,7 +717,7 @@ pub mod pallet {
 
                 ContractEraStake::<T>::insert(contract_id.clone(), current_era, staking_info);
 
-                Self::deposit_event(Event::<T>::RewardRestaked(
+                Self::deposit_event(Event::<T>::RewardAndRestake(
                     staker.clone(),
                     contract_id.clone(),
                     staker_reward,
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -1031,7 +1031,7 @@ pub mod pallet {
         }
 
         // `true` if all the conditions for restaking the reward have been met, `false` otherwise
-        fn should_restake_reward(
+        pub(crate) fn should_restake_reward(
             reward_destination: RewardDestination,
             contract_id: &T::SmartContract,
             latest_staked_value: BalanceOf<T>,
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -504,9 +504,19 @@ pub mod pallet {
             );
 
             // Increment ledger and total staker value for contract. Overflow shouldn't be possible but the check is here just for safety.
-            Self::stake(value_to_stake, &mut staker_info);
-            Self::add_ledger_locked(&staker, &mut ledger, value_to_stake);
-            Self::update_contract_era_stake(&mut staking_info, &contract_id, value_to_stake);
+            ensure!(
+                Self::stake(value_to_stake, &mut staker_info).is_ok(),
+                Error::<T>::UnexpectedStakeInfoEra
+            );
+            ensure!(
+                Self::add_ledger_locked(&staker, &mut ledger, value_to_stake).is_ok(),
+                ArithmeticError::Overflow
+            );
+            ensure!(
+                Self::update_contract_era_stake(&mut staking_info, &contract_id, value_to_stake)
+                    .is_ok(),
+                ArithmeticError::Overflow
+            );
             Self::update_staker_info(&staker, &contract_id, staker_info.clone());
 
             Self::deposit_event(Event::<T>::BondAndStake(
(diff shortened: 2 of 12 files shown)


frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -890,6 +862,63 @@ pub mod pallet {
...
+            ensure!(
+                staker_info.len() <= T::MaxEraStakeValues::get(),
+                Error::<T>::TooManyEraStakeValues
+            );
+            ensure!(
+                staker_info.latest_staked_value() >= T::MinimumStakingAmount::get(),
+                Error::<T>::InsufficientValue,
+            );
+
+            let mut ledger = Self::ledger(&staker);
```
codespool, line 902: pass instead of reading again

hoonsubin:
Sorry! I didn't mean to remove the branch! I think I made a mistake when removing my local branch. Please reopen this!

frame/dapps-staking/src/benchmarking.rs
```diff
@@ -230,6 +230,16 @@ benchmarks! {
...
+    enable_compound_staking {
+        initialize::<T>();
+
+        let staker: T::AccountId = whitelisted_caller();
+        let option = RewardHandling::OnlyPayout;
+    }: _(RawOrigin::Signed(staker.clone()), option)
+    verify {
+        assert_last_event::<T>(Event::<T>::RewardHandlingChange(staker, option).into());
+    }
+
```
Dinonard, line 279: Good!

You will also need to create separate `claim_staker` benchmarking for both cases: using compounding and not using it.

You can find examples in existing substrate pallets.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -864,6 +884,22 @@ pub mod pallet {
             PreApprovalIsEnabled::<T>::put(enabled);
             Ok(().into())
         }
+
+        #[pallet::weight(T::WeightInfo::enable_compound_staking())]
+        pub fn enable_compound_staking(
```
Dinonard: Perhaps `set_reward_destination` is more general?

Btw, enum has been renamed to `RewardDestination` in latest PR version.

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -419,13 +419,23 @@ pub(crate) fn assert_claim_staker(claimer: AccountId, contract_id: &MockSmartCon
...
+            calculated_reward,
+        )))
+    } else {
+        System::assert_last_event(mock::Event::DappsStaking(Event::Reward(
+            claimer,
+            contract_id.clone(),
+            claim_era,
+            calculated_reward,
+        )));
+    }
```
Dinonard, line 456: I believe you should also expand the checked storage values AFTER the call.

E.g. locked/staked values should be increased for 3 separate storage entries.
In case compounding isn't used, they should remain the same.

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -419,13 +419,23 @@ pub(crate) fn assert_claim_staker(claimer: AccountId, contract_id: &MockSmartCon
...
     ));
-    System::assert_last_event(mock::Event::DappsStaking(Event::Reward(
-        claimer,
-        contract_id.clone(),
-        claim_era,
-        calculated_reward,
-    )));
 
+    if init_state.ledger.reward_handling == RewardHandling::PayoutAndStake
+        && !init_state.staker_info.latest_staked_value().is_zero()
```
Dinonard: Can this be a helper method?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -872,6 +908,25 @@ pub mod pallet {
             T::PalletId::get().into_account()
         }
 
+        fn update_staking_values(
```
Dinonard: This is just my opinion, but I'm not very fond of this utility method.

It does too much different things, imo.
Updates 3 distinct values, but only one gets written into storage.
I understand that the purpose is to avoid code duplication but perhaps it's fine to have these few lines duplicated.

What do you think?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -741,15 +735,41 @@ pub mod pallet {
...
-            Self::deposit_event(Event::<T>::Reward(
-                staker.clone(),
-                contract_id.clone(),
-                era,
-                staker_reward,
-            ));
+            let mut ledger = Self::ledger(&staker);
 
+            if ledger.reward_handling == RewardHandling::PayoutAndStake
+                && !staker_info.latest_staked_value().is_zero()
```
Dinonard: Come to think of it, there's also a third-scenario: contract has been unregistered.

Perhaps this should be a utility method?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -741,15 +735,41 @@ pub mod pallet {
...
+            {
+                staker_info
+                    .stake(current_era, staker_reward)
+                    .map_err(|_| Error::<T>::UnexpectedStakeInfoEra)?;
```
Dinonard, line 708: I've been thinking about this and believe we have to handle scenario when number of `EraStake` values in `stakes` vector is exceeded.

### Example
// <era, staked amount>
**stakers** = [`<1, 100>`, `<95, 300>`]

This tells us that staker had staked **100** in era **1** and then increased his stake to **300** in era **95**.
NO era rewards have been claimed.

No consider a scenario where staker decides to do the following:
* Claim in era **96** with compounding rewards. **stakers** becomes [<...>, <...>, `<96, 301>`]
* Claim in era **97** with compounding rewards. **stakers** becomes [<...>, <...>, `<96, 301>`, `<97, 302>`]
* ...

By doing this repeatedly, vec can grow to uncontrolled size.

### Solution
* Add a check for `stakers` vector size.
* I will modify the original PR to allow `bond_and_stake` and `unbond_and_unstake` to only allow vec to grow up to `MaxEraStakeValues - 1`.
* This will allow batch call of `claim_staker` to claim multiple eras at once.
* If staker gets stuck, they can always switch **compounding rewards** off and reduce the size of `stakers` vec.

What do you think about that?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -741,15 +735,41 @@ pub mod pallet {
...
+                    contract_id.clone(),
+                    staker_reward,
+                ));
+            } else {
+                Self::deposit_event(Event::<T>::Reward(
+                    staker.clone(),
+                    contract_id.clone(),
+                    era,
+                    staker_reward,
+                ));
```
Dinonard: It is important we still emit the old event as before.

External services use it to calculate how much rewards were earned by a particular account.
If we change this, they all need to update too and we want to avoid that.

You can just emit two separate events.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -723,7 +717,7 @@ pub mod pallet {
             let current_era = Self::current_era();
             ensure!(era < current_era, Error::<T>::EraOutOfBounds);
 
-            let staking_info = Self::staking_info(&contract_id, era);
+            let mut staking_info = Self::staking_info(&contract_id, era);
             let reward_and_stake =
```
Dinonard, line 689: You already read `EraInfo` here so you shouldn't do it again in the utility method `update_staking_values`

frame/dapps-staking/src/tests.rs
```diff
@@ -1289,6 +1289,51 @@ fn claim_after_unregister_is_ok() {
...
+        let start_era = DappsStaking::current_era();
+        assert_register(developer, &contract_id);
+        let stake_value = 100;
+        assert_bond_and_stake(staker, &contract_id, stake_value);
+
+        advance_to_era(start_era + 1);
+        assert_ok!(DappsStaking::enable_compound_staking(
+            Origin::signed(staker),
+            RewardHandling::OnlyPayout
+        ));
```
Dinonard: I suggest you add an `assert_` utility function for this where you assert the event and init/final state.

frame/dapps-staking/src/tests.rs
```diff
@@ -1289,6 +1289,51 @@ fn claim_after_unregister_is_ok() {
...
+        let developer = 1;
+        let staker = 2;
+        let contract_id = MockSmartContract::Evm(H160::repeat_byte(0x01));
+
+        let start_era = DappsStaking::current_era();
+        assert_register(developer, &contract_id);
+        let stake_value = 100;
+        assert_bond_and_stake(staker, &contract_id, stake_value);
+
+        advance_to_era(start_era + 1);
```
Dinonard, line 1489: Perhaps a few inline comments describing what's being done would be helpful?

frame/dapps-staking/src/tests.rs
```diff
@@ -1289,6 +1289,51 @@ fn claim_after_unregister_is_ok() {
...
+
+        assert_unbond_and_unstake(staker, &contract_id, stake_value);
+
+        assert_claim_staker(staker, &contract_id);
+    })
+}
```
Dinonard, line 1500: How about covering the new extrinsic call with UT?

Also, these scenarios should be covered:
* switch between different reward payouts withing one UT and verify it works
* Produce a scenario where `stakers` vec size would go out of limit
  * verify that if user switches off compounding rewards, claim can once again work

Dinonard (commented):
Good progress! :+1:

frame/dapps-staking/src/pallet/mod.rs
codespool: agreed, I imagined it differently when I started, but now it's just proxying a call to two mutations.
removed.

frame/dapps-staking/src/pallet/mod.rs
codespool: removed the utility method.

frame/dapps-staking/src/pallet/mod.rs
codespool: Makes sense. I added the check.

Dinonard:
Please rebase on top of latesat `polkadot-v0.9.17` and I'll check the code.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -253,6 +253,10 @@ pub mod pallet {
         Reward(T::AccountId, T::SmartContract, EraIndex, BalanceOf<T>),
         /// Maintenance mode has been enabled or disabled
         MaintenanceMode(bool),
+        /// Claimed staker reward restaked
+        RewardAndRestake(T::AccountId, T::SmartContract, BalanceOf<T>),
```
Dinonard: Should we just reuse `BondAndStake` event?

After all, it does convey the same information.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -582,7 +580,7 @@ pub mod pallet {
 
             Self::update_ledger(&staker, ledger);
             Self::update_staker_info(&staker, &contract_id, staker_info);
-            ContractEraStake::<T>::insert(&contract_id, current_era, staking_info);
+            ContractEraStake::<T>::insert(contract_id.clone(), current_era, staking_info);
```
Dinonard: Is this change needed?
Why can't reference be used?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -751,6 +750,44 @@ pub mod pallet {
             let staker_reward =
                 Perbill::from_rational(staked, staking_info.total) * stakers_joint_reward;
 
+            let mut ledger = Self::ledger(&staker);
+
+            if Self::should_restake_reward(
+                ledger.reward_destination,
+                &contract_id,
```
Dinonard: You shouldn't read `DAppInfo` twice.

It's already read on L734.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -751,6 +750,44 @@ pub mod pallet {
...
+                staker_info.latest_staked_value(),
+            ) {
+                // There must be one slot left for restaking
+                ensure!(
+                    staker_info.len() < T::MaxEraStakeValues::get(),
+                    Error::<T>::TooManyEraStakeValues
+                );
```
Dinonard: This isn't correct - you should do the check with `<=` AFTER calling `stake` on `staker_info` struct.

This is important since in a scenario like this:
```
current_era = 15;
MaxEraStakeValues = 3;
stakes = [<1000, 5>, <2000, 10>]

UI prepares a batch call of claims (e.g. let's say there are 10 claim_staker calls).
1. [<1000, 6>, <2000, 10>, <2001, 15>]
2. ERROR
3. ERROR
...
10. ERROR
```
But it's not a problem since an existing `EraStake` value will be modified.

Please also add UT for this scenario.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -751,6 +750,44 @@ pub mod pallet {
...
+                Self::update_ledger(&staker, ledger);
+
+                ContractEraStake::<T>::insert(contract_id.clone(), current_era, staking_info);
+
+                Self::deposit_event(Event::<T>::RewardAndRestake(
+                    staker.clone(),
+                    contract_id.clone(),
+                    staker_reward,
+                ));
+            }
```
Dinonard: You cannot do this here since you have code which can possibly fail on L792.

You should probably move that part BEFORE this `if` clause.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -760,16 +797,19 @@ pub mod pallet {
...
                 contract_id.clone(),
                 era,
                 staker_reward,
             ));
-
-            Ok(().into())
+            Self::update_staker_info(&staker, &contract_id, staker_info);
+            Ok(Some(
+                T::WeightInfo::claim_staker_with_restake()
+                    .max(T::WeightInfo::claim_staker_without_restake()),
```
Dinonard: At this point you should return the correct one which was used.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -760,16 +797,19 @@ pub mod pallet {
...
 
             Self::deposit_event(Event::<T>::Reward(
                 staker.clone(),
                 contract_id.clone(),
                 era,
                 staker_reward,
             ));
-
-            Ok(().into())
+            Self::update_staker_info(&staker, &contract_id, staker_info);
```
Dinonard, line 750: Not sure I understand why the move is needed - can you elaborate?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -253,6 +253,10 @@ pub mod pallet {
         Reward(T::AccountId, T::SmartContract, EraIndex, BalanceOf<T>),
         /// Maintenance mode has been enabled or disabled
         MaintenanceMode(bool),
+        /// Claimed staker reward restaked
+        RewardAndRestake(T::AccountId, T::SmartContract, BalanceOf<T>),
+        /// Reward handling modified
+        RewardDestinationSet(T::AccountId, RewardDestination),
```
Dinonard: Can we use just `RewardDestination`?

Imo, it's more concise and conveys the same info.

frame/dapps-staking/src/tests.rs
```diff
@@ -1360,6 +1360,123 @@ fn claim_after_unregister_is_ok() {
         }
     })
 }
+#[test]
```
Dinonard, line 1451: Please separate it from above UT with new line.

frame/dapps-staking/src/tests.rs
```diff
@@ -1360,6 +1360,123 @@ fn claim_after_unregister_is_ok() {
...
+
+        // there's one more stake spot left by bond_and_stake
+        // it's filled by claiming and restaking once
+        assert_claim_staker(staker_id, &contract_id);
+
+        // claiming should not work because it can't stake any more
+        assert_noop!(
+            DappsStaking::claim_staker(Origin::signed(staker_id), contract_id),
+            Error::<TestRuntime>::TooManyEraStakeValues
+        );
```
Dinonard: It should work :)

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -394,43 +394,93 @@ pub(crate) fn assert_withdraw_unbonded(staker: AccountId) {
...
         Origin::signed(claimer),
         contract_id.clone(),
     ));
+
+    let final_state_claim_era = MemorySnapshot::all(claim_era, contract_id, claimer);
+    let final_state_current_era = MemorySnapshot::all(current_era, contract_id, claimer);
+
+    // If there's less than 2 events, it should fail anyway, so panic as acceptable
+    let events = dapps_staking_events();
+    let second_last_event = &events[events.len() - 2];
```
Dinonard: I'd recommend you clear all events prior to calling `claim_staker` to ensure you don't have any leftovers from old calls.

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -394,43 +394,93 @@ pub(crate) fn assert_withdraw_unbonded(staker: AccountId) {
...
+
+    if DappsStaking::should_restake_reward(
+        init_state_claim_era.ledger.reward_destination,
+        contract_id,
+        init_state_claim_era.staker_info.latest_staked_value(),
+    ) {
+        assert_eq!(
+            second_last_event.clone(),
+            Event::<TestRuntime>::RewardAndRestake(claimer, contract_id.clone(), calculated_reward,)
```
Dinonard: You're only checking whether one event is correct in this case.

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -394,43 +394,93 @@ pub(crate) fn assert_withdraw_unbonded(staker: AccountId) {
...
+        );
+        assert_eq!(
+            init_state_current_era.era_info.locked + calculated_reward,
+            final_state_current_era.era_info.locked
+        );
+    } else {
+        assert_eq!(
+            init_state_claim_era.free_balance + calculated_reward,
+            final_state_claim_era.free_balance
+        );
```
Dinonard: Isn't this check also applicable if `if` clause is `true`?

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -394,43 +394,93 @@ pub(crate) fn assert_withdraw_unbonded(staker: AccountId) {
...
     ));
+
+    let final_state_claim_era = MemorySnapshot::all(claim_era, contract_id, claimer);
+    let final_state_current_era = MemorySnapshot::all(current_era, contract_id, claimer);
+
+    // If there's less than 2 events, it should fail anyway, so panic as acceptable
+    let events = dapps_staking_events();
+    let second_last_event = &events[events.len() - 2];
+
+    if DappsStaking::should_restake_reward(
```
Dinonard, line 443: Since this is `pub`, you should also probably add a few UTs for it.

Also, this function has become quite large - maybe having separate helper functions for case when no restake is done and when it is would help increase readability.

Dinonard (commented):
Please add some comments to functions/methods that are missing them.

frame/dapps-staking/src/pallet/mod.rs
codespool: Yes, makes sense.

frame/dapps-staking/src/pallet/mod.rs
codespool: Sure we can.. get and set sufixes are just old habits :D

frame/dapps-staking/src/pallet/mod.rs
codespool: Good catch, thanks!

frame/dapps-staking/src/pallet/mod.rs
codespool: in case of a batch call, only the last item would be mutated, right?
so, something like:
```rust
ensure!(
    staker_info.len() < T::MaxEraStakeValues::get() 
    || staker_info.latest_staked_era == current_era,
    Error::<T>::TooManyEraStakeValues
);
```
should make the result:
```
1. [<1000, 6>, <2000, 10>, <2001, 15>] 
2. [<1000, 7>, <2000, 10>, <2002, 15>] 
3. [<1000, 8>, <2000, 10>, <2003, 15>]
4. [<1000, 9>, <2000, 10>, <2004, 15>]
5. [<2000, 10>, <2005, 15>]
6. [<2000, 11>, <2007, 15>]
7. [<2000, 12>, <2009, 15>]
8. [<2000, 13>, <2011, 15>]
9.[<2000,14>,<2013,15>]
10.[<2015, 15>]
```

Should I add the `last_staked_era` function and check for it here?

frame/dapps-staking/src/tests.rs
codespool: Should it? The call before has filled the `stakes` vector, and it shouldn't be possible to claim with reward_destination set to `stakeBalance`, which is changed to `freeBalance` in the next call.

frame/dapps-staking/src/pallet/mod.rs
codespool: it's not if I call update_staker_info first :+1:

frame/dapps-staking/src/pallet/mod.rs
codespool: I wasn't clear on how to do it from the examples. Should I just `if` the return value?

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Yes, that's correct.
I gave this batch example just for the demonstration of how it should work.
You don't need to use `last_staked_era` for this check. :)

Just use the exact same check as in `bond_and_stake` and `unbond_and_unstake`.
With a small difference: use `<=` instead of `<`.

Just move L761 code below L766.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Sure!

I guess if you store the `bool` of from L755 check, it becomes a simple `if` :)

frame/dapps-staking/src/tests.rs
Dinonard: I think this is related to: https://github.com/AstarNetwork/astar-frame/pull/14#discussion_r830860865

Since era remains the same, it's not possible for consecutive calls of `claim_staker` to increase the size of the vector.

frame/dapps-staking/src/pallet/mod.rs
codespool: I extracted the condition and moved it before the call to `T::Currency::withdraw` because the body of the `if` clause also had possible errors.

frame/dapps-staking/src/testing_utils.rs
codespool: yes, it's a mistake, it should check free balance in the `current_era` is higher after claim without compounding

frame/dapps-staking/src/testing_utils.rs
codespool: yes, the last event is asserted regardless of the `if` clause (`Reward` should be dispatched in both cases)

frame/dapps-staking/src/benchmarking.rs
```diff
@@ -195,7 +195,7 @@ benchmarks! {
         assert_last_event::<T>(Event::<T>::Withdrawn(staker, unstake_amount).into());
     }
 
-    claim_staker {
+    claim_staker_with_restake {
```
Dinonard, line 207: One small change I'd recommend here: set the restake option explicity.

If by any chance we modify the default value in the future, this won't need update.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -562,15 +564,9 @@ pub mod pallet {
...
-                .locked
-                .checked_add(&value_to_stake)
-                .ok_or(ArithmeticError::Overflow)?;
-            staking_info.total = staking_info
-                .total
-                .checked_add(&value_to_stake)
-                .ok_or(ArithmeticError::Overflow)?;
+            // Increment ledger and total staker value for contract.
+            ledger.locked = ledger.locked.saturating_add(value_to_stake);
+            staking_info.total = staking_info.total.saturating_add(value_to_stake);
```
Dinonard, line 515: Interesting enough, I've found a neat way how we can inject `T: Config` as dependency into these structs without causing us too much problems.

So basically, we could have this method return an `Result<_,_>` and just use `?` operator to either get the ok value or return error.

Not for this commit though :)

If you're interested, check `pallet-staking` in substrate repo.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -751,6 +748,25 @@ pub mod pallet {
...
+                dapp_info.state,
+                staker_info.latest_staked_value(),
+            );
+
+            if should_restake_reward {
+                staker_info
+                    .stake(current_era, staker_reward)
+                    .map_err(|_| Error::<T>::UnexpectedStakeInfoEra)?;
+
+                ensure!(
```
Dinonard, line 712: Would an inline comment here be a good idea?

Since we have a different check here and in `bond_and_stake` + `unbond_and_unstake` calls.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -759,17 +775,39 @@ pub mod pallet {
...
-                staker_reward,
-            ));
-
-            Ok(().into())
+            Ok(Some(if should_restake_reward {
+                T::WeightInfo::claim_staker_with_restake()
+            } else {
+                T::WeightInfo::claim_staker_without_restake()
+            })
+            .into())
```
Dinonard, line 758: Nice :)

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -394,43 +394,68 @@ pub(crate) fn assert_withdraw_unbonded(staker: AccountId) {
...
+        &init_state_claim_era.era_info,
+    );
 
-    let (claim_era, staked) = init_state.staker_info.clone().claim();
+    let (claim_era, staked) = init_state_claim_era.staker_info.clone().claim();
     assert!(claim_era > 0); // Sanity check - if this fails, method is being used incorrectly
 
     // Cannot claim rewards post unregister era, this indicates a bug!
-    if let DAppState::Unregistered(unregistered_era) = init_state.dapp_info.state {
+    if let DAppState::Unregistered(unregistered_era) = init_state_claim_era.dapp_info.state {
```
Dinonard, line 418: I missed this before. Good thing you noticed it!

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -446,6 +471,41 @@ pub(crate) fn assert_claim_staker(claimer: AccountId, contract_id: &MockSmartCon
     assert_eq!(issuance_before_claim, issuance_after_claim);
 }
 
+// assert staked and locked states depending on should_restake_reward
+// returns should_restake_reward result so further checks can be made
+fn assert_restake_reward(
```
Dinonard, line 485: Great that we have a util methods for this.

Can we keep them separated though?
We use `assert_` style methods as non-return functions which cause a panic in case some expected condition isn't met.

I'd recommend just removing `-> bool`, and calling `should_restake_reward` explicitly where you need it.

frame/dapps-staking/src/tests.rs
```diff
@@ -1361,6 +1361,124 @@ fn claim_after_unregister_is_ok() {
...
+        let contract_id = MockSmartContract::Evm(H160::repeat_byte(0x01));
+        let start_era = DappsStaking::current_era();
+        assert_register(developer, &contract_id);
+
+        // stake some tokens and wait for an era
+        let stake_value = 100;
+        assert_bond_and_stake(staker, &contract_id, stake_value);
+        advance_to_era(start_era + 1);
+
+        // ensure reward_destination is set tot StakeBalance
```
Dinonard: You should disable spell checker German settings :grin:

frame/dapps-staking/src/tests.rs
```diff
@@ -1361,6 +1361,124 @@ fn claim_after_unregister_is_ok() {
...
+
+        // disable compounding mode, wait 3 eras
+        assert_set_reward_destination(staker, RewardDestination::FreeBalance);
+        advance_to_era(start_era + 3);
+        // ensure staker can claim rewards to wallet
+        assert_claim_staker(staker, &contract_id);
+
+        // enable compounding mode, wait 3 eras
+        assert_set_reward_destination(staker, RewardDestination::StakeBalance);
+        advance_to_era(start_era + 3);
```
Dinonard: I'm not sure I understand why this part is needed - could you explain?

Could we just advance two eras, call `claim_staker`, change reward destination, call `claim_staker` again?

frame/dapps-staking/src/tests.rs
```diff
@@ -1361,6 +1361,124 @@ fn claim_after_unregister_is_ok() {
...
+fn claiming_when_stakes_full_without_compounding_is_ok() {
+    ExternalityBuilder::build().execute_with(|| {
+        initialize_first_block();
+
+        let staker_id = 1;
+        let contract_id = MockSmartContract::Evm(H160::repeat_byte(0x01));
+        // Insert a contract under registered contracts.
+        assert_register(10, &contract_id);
+
+        // Stake with MAX_NUMBER_OF_STAKERS - 1 on the same contract. It must work.
```
Dinonard: MAX_ERA_STAKE_VALUES typo

frame/dapps-staking/src/tests.rs
```diff
@@ -1361,6 +1361,124 @@ fn claim_after_unregister_is_ok() {
...
+        }
+
+        // Make sure reward_destination is set to StakeBalance
+        assert_set_reward_destination(staker_id, RewardDestination::StakeBalance);
+
+        // claim and restake once, so there's a claim record for the for the current era in the stakes vec
+        assert_claim_staker(staker_id, &contract_id);
+
+        // making another gap in eras and trying to claim and restake would exceed MAX_ERA_STAKE_VALUES
+        advance_to_era(MAX_ERA_STAKE_VALUES * 5);
```
Dinonard: Maybe jusy `current_era() +1`? :)

The current argument looks out of place.

frame/dapps-staking/src/weights.rs
```diff
@@ -29,7 +31,7 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: DappsStaking RegisteredDapps (r:1 w:1)
 	// Storage: DappsStaking PreApprovalIsEnabled (r:1 w:0)
 	fn register() -> Weight {
-		(32_139_000 as Weight)
+		(60_000_000 as Weight)
```
Dinonard: We'll have to see about all these changes...we have a problem with current weights.

I ran them on my laptop which is more powerful than the referent machine which should be used to measure weights. So currently, all of our weights are a bit too small!

But it seems your execution increased weights quite a bit - so maybe your machine is closer to the referent machine specs. Did you ensure nothing else is running when you ran benchmarks?

Anyhow, you could try benchmarking `pallet-balances` to generate your personal weights file and then compare it to the official weights in substrate repo. That should tell you in a sense how close your machine to referent machine is.

frame/dapps-staking/src/weights.rs
```diff
@@ -122,23 +139,30 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-		(10_970_000 as Weight)
+		(18_000_000 as Weight)
 			.saturating_add(T::DbWeight::get().reads(1 as Weight))
 			.saturating_add(T::DbWeight::get().writes(1 as Weight))
 	}
+	// Storage: DappsStaking PalletDisabled (r:1 w:0)
+	// Storage: DappsStaking Ledger (r:1 w:1)
+	fn set_reward_destination() -> Weight {
+		(24_000_000 as Weight)
+			.saturating_add(T::DbWeight::get().reads(2 as Weight))
```
Dinonard, line 163: Can this be optimized?
Do you need 2 reads or is 1 enough?

Dinonard (commented):
Very good! :+1: 

A few small comments but otherwise I think you're done.

frame/dapps-staking/src/pallet/mod.rs
codespool: nice! will take a look 👍

frame/dapps-staking/src/testing_utils.rs
codespool: well, yeah, it took me a while to figure out snapshots capture a specific era :)
I created two separate snapshots, one for `claim_era`, one for `current_era`, do you mind double checking the logic for the assertions now, in case I messed something up?

frame/dapps-staking/src/tests.rs
codespool: hahah, I think it ignores the words with <= 3 characters 😄

frame/dapps-staking/src/tests.rs
codespool: oh, it's actually a typo (both cases advancing to `start_era + 3`, it should be +1 and +2).
Idea was to test both cases, regardless of the default `RewardDestination`.
So, 
set destination to `FreeBalance` --> advance era --> claim --> set destination to `StakeBalance` --> advance era --> claim
 Advancing eras in between claims so there's something to claim.

frame/dapps-staking/src/weights.rs
codespool: not sure what you mean, 1 read is to check is pallet disabled, the other is to get ledger for the origin account.
Can I drop one of them?

frame/dapps-staking/src/weights.rs
codespool: I tried benchmarking `pallet_balances`, and I also get the double that of the official repo.
I'm running it on macbook pro, 2,3 GHz 8-Core Intel Core i9, 16 GB 2667 MHz DDR4, which is considerably weaker than the reference machine.

Dinonard: Oh!

In that case, perhaps it's better I run benchmarks on my machine just so we keep it consistent.

frame/dapps-staking/src/testing_utils.rs
Dinonard: I did during review :)

It looks good to me!

frame/dapps-staking/src/weights.rs
Dinonard: Oh my bad, it skipped my mind that it's a struct we need to modify - not just a simple entry.

Disregard the comment.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -967,7 +969,7 @@ pub mod pallet {
             T::PalletId::get().into_account()
         }
 
-        /// `Ok` if pallet disabled for maintenance, `Err` otherwise
+        /// `Err` if pallet disabled for maintenance, `Ok` otherwise
```
Dinonard, line 922: Good catch :see_no_evil:

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -473,17 +479,12 @@ pub(crate) fn assert_claim_staker(claimer: AccountId, contract_id: &MockSmartCon
 
 // assert staked and locked states depending on should_restake_reward
 // returns should_restake_reward result so further checks can be made
-fn assert_restake_reward(
-    init_state: MemorySnapshot,
-    final_state: MemorySnapshot,
-    reward: u128,
-) -> bool {
-    let should_restake_reward = DappsStaking::should_restake_reward(
+fn assert_restake_reward(init_state: &MemorySnapshot, final_state: &MemorySnapshot, reward: u128) {
```
Dinonard: Better to use `Balance` as type for reward.

Dinonard (commented):
LGTM :+1: 

This is an approval basically, we just have to do it in order as discussed on Slack.

Well done! :smile:

codespool:
@Dinonard rebased 👍

Dinonard:
@codespool let's get this merged and deployed on Shibuya :+1: 

Can you please prepare a PR in `Astar` repo which adds compounding rewards to `local` and `Shibuya`?
You can just use this branch as dependency.

Once you have that, I can run benchmarks and give you the values so you can add them into this PR.

frame/dapps-staking/src/weights.rs
```diff
@@ -256,6 +295,12 @@ impl WeightInfo for () {
...
 		(10_970_000 as Weight)
 			.saturating_add(RocksDbWeight::get().reads(1 as Weight))
+	}
+	// Storage: DappsStaking PalletDisabled (r:1 w:0)
+	// Storage: DappsStaking Ledger (r:1 w:1)
+	fn set_reward_destination() -> Weight {
+		(24_000_000 as Weight)
+			.saturating_add(RocksDbWeight::get().reads(2 as Weight))
 			.saturating_add(RocksDbWeight::get().writes(1 as Weight))
 	}
```
Dinonard, line 306:
```Rust
	// Storage: DappsStaking PalletDisabled (r:1 w:0)
	// Storage: DappsStaking GeneralStakerInfo (r:1 w:1)
	// Storage: DappsStaking RegisteredDapps (r:1 w:0)
	// Storage: DappsStaking CurrentEra (r:1 w:0)
	// Storage: DappsStaking ContractEraStake (r:1 w:1)
	// Storage: DappsStaking GeneralEraInfo (r:2 w:1)
	// Storage: DappsStaking Ledger (r:1 w:1)
	// Storage: Balances Locks (r:1 w:1)
	// Storage: System Account (r:1 w:1)
	fn claim_staker_with_restake() -> Weight {
		(78_506_000 as Weight)
			.saturating_add(T::DbWeight::get().reads(10 as Weight))
			.saturating_add(T::DbWeight::get().writes(6 as Weight))
	}
	// Storage: DappsStaking PalletDisabled (r:1 w:0)
	// Storage: DappsStaking GeneralStakerInfo (r:1 w:1)
	// Storage: DappsStaking RegisteredDapps (r:1 w:0)
	// Storage: DappsStaking CurrentEra (r:1 w:0)
	// Storage: DappsStaking ContractEraStake (r:1 w:0)
	// Storage: DappsStaking GeneralEraInfo (r:1 w:0)
	// Storage: DappsStaking Ledger (r:1 w:0)
	fn claim_staker_without_restake() -> Weight {
		(56_575_000 as Weight)
			.saturating_add(T::DbWeight::get().reads(7 as Weight))
			.saturating_add(T::DbWeight::get().writes(1 as Weight))
	}
	// Storage: DappsStaking PalletDisabled (r:1 w:0)
	// Storage: DappsStaking Ledger (r:1 w:1)
	fn set_reward_destination() -> Weight {
		(15_489_000 as Weight)
			.saturating_add(T::DbWeight::get().reads(2 as Weight))
			.saturating_add(T::DbWeight::get().writes(1 as Weight))
	}
```

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -845,7 +885,24 @@ pub mod pallet {
...
+        #[pallet::weight(T::WeightInfo::set_reward_destination())]
+        pub fn set_reward_destination(
+            origin: OriginFor<T>,
+            reward_destination: RewardDestination,
+        ) -> DispatchResultWithPostInfo {
+            Self::ensure_pallet_enabled()?;
+            let staker = ensure_signed(origin)?;
+
+            Ledger::<T>::mutate(&staker, |ledger| {
+                ledger.reward_destination = reward_destination;
```
Dinonard: Just noticed that this will leave a DB entry even if an account isn't staking.
This shouldn't be the case since we cleanup empty ledgers.

You should ensure that the ledger isn't empty before mutating it.
Add a UT for the scenario too.

Dinonard (changes requested):

frame/dapps-staking/src/pallet/mod.rs
codespool: good catch 👍

frame/dapps-staking/src/weights.rs
codespool: thnx!

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -898,6 +900,11 @@ pub mod pallet {
             Self::ensure_pallet_enabled()?;
             let staker = ensure_signed(origin)?;
 
+            ensure!(
+                Ledger::<T>::contains_key(&staker),
+                Error::<T>::NotActiveStaker
+            );
```
Dinonard: We should keep it to 1 read and 1 write - `contains_key` counts as a read now.

I'd suggest to read the value from the DB, check it using the `is_empty()` method and then insert the modified value.

frame/dapps-staking/src/weights.rs
Dinonard: Please only modify the 3 benchmarks you impacted, let's leave the other numbers as they are for now.

Dinonard (approved):

Dinonard:
@codespool please fix the formatting and let's merge.

PR #17, opened 2022-02-22
Individual claim - claim_dapp without era param

Dinonard:
**Pull Request Summary**

In order to simplify frontend implementation and backend storage handling,
it is no longer needed to specify which era is being claimed when claiming dapp rewards.
As a rule, oldest available era will always be claimed.

* Modified `claim_dapp` to support claiming without era param
* Modified benchmark & try-runtime code
* Modified test utils
* Modified and expanded UTs

**TODOs**
- [x] implement new logic
- [x] adjust & expand UTs
- [ ] live test

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


frame/dapps-staking/src/lib.rs
```diff
@@ -59,40 +87,29 @@ impl Default for Forcing {
...
+    pub locked: Balance,
 }
 
 /// Used to split total EraPayout among contracts.
 /// Each tuple (contract, era) has this structure.
 /// This will be used to reward contracts developer and his stakers.
-#[derive(Clone, PartialEq, Encode, Decode, RuntimeDebug, TypeInfo)]
-pub struct EraStakingPoints<AccountId: Ord, Balance: HasCompact> {
+#[derive(Clone, PartialEq, Encode, Decode, Default, RuntimeDebug, TypeInfo)]
+pub struct EraStakingPoints<Balance: HasCompact> {
```
akru, line 106: This naming a bit confusing because of correlation with EraInfo. May be just `StakingPoints` or `DAppStakingPoints` will looks better.

frame/dapps-staking/src/lib.rs
```diff
@@ -101,11 +118,154 @@ impl<AccountId: Ord, Balance: HasCompact + Default> Default
...
+    V3_0_0,
 }
 
 impl Default for Version {
     fn default() -> Self {
-        Version::V1_0_0
+        Version::V2_0_0
+    }
+}
+
```
akru, line 129: Please add description of this structure.

frame/dapps-staking/src/lib.rs
```diff
@@ -101,11 +118,154 @@ impl<AccountId: Ord, Balance: HasCompact + Default> Default
...
+    era: EraIndex,
+}
+
+impl<Balance: AtLeast32BitUnsigned + Copy> EraStake<Balance> {
+    fn new(staked: Balance, era: EraIndex) -> Self {
+        Self { staked, era }
+    }
+}
+
+/// Contains information about how much was staked in each era
```
akru, line 146: It's definitely good idea to move most sensitive staking actions into one struct implementation, unfortunately, motivation of used algorithm is a bit unclear. Could you add more information into structure description to make it clear, preferred to have some examples.

frame/dapps-staking/src/lib.rs
```diff
@@ -101,11 +118,154 @@ impl<AccountId: Ord, Balance: HasCompact + Default> Default
...
+    /// number of `EraStake` chunks
+    fn len(&self) -> u32 {
+        self.stakes.len() as u32
+    }
+
+    /// Stakes some value in the specified era.
+    /// User should ensure that given era is either equal or greater than the
+    /// latest available era in the staking info.
+    fn stake(&mut self, current_era: EraIndex, value: Balance) -> Result<(), &str> {
+        if self.stakes.is_empty() {
```
akru, line 168: What's about `if let Some(x) = self.stakes.last()`?

frame/dapps-staking/src/lib.rs
```diff
@@ -101,11 +118,154 @@ impl<AccountId: Ord, Balance: HasCompact + Default> Default
...
+        } else {
+            let era_stake = self.stakes.last().unwrap(); // exists if vec not empty
+            if era_stake.era > current_era {
+                return Err("Unexpected era".into());
+            }
+
+            let new_stake_value = era_stake.staked.saturating_add(value);
+
+            if current_era == era_stake.era {
+                *self.stakes.last_mut().unwrap() = EraStake::new(new_stake_value, current_era)
```
akru, line 179: We can escape of using `unwrap()` here, look comment above.

frame/dapps-staking/src/lib.rs
```diff
@@ -101,11 +118,154 @@ impl<AccountId: Ord, Balance: HasCompact + Default> Default
...
+
+        let era_stake = self.stakes.last().unwrap(); // not empty so it exists
+        if era_stake.era > current_era {
+            return Err("Unexpected era".into());
+        }
+
+        let new_stake_value = era_stake.staked.saturating_sub(value);
+
+        if current_era == era_stake.era {
+            *self.stakes.last_mut().unwrap() = EraStake::new(new_stake_value, current_era)
```
akru, line 204: Let's use the same `if let` here to escape of using `unwrap` often.

frame/dapps-staking/src/lib.rs
```diff
@@ -208,12 +368,39 @@ where
     }
 }
 
+/// Instruction on how to handle reward payout for stakers.
+/// In order to make staking more competitive, majority of stakers will want to
+/// automatically restake anything they earn.
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub enum RewardHandling {
```
akru, line 375: What's about `RewardDestination`? And options: `Balance` or `Stake`.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -161,13 +166,18 @@ pub mod pallet {
     #[pallet::storage]
     #[pallet::getter(fn registered_developer)]
     pub(crate) type RegisteredDapps<T: Config> =
-        StorageMap<_, Blake2_128Concat, T::SmartContract, T::AccountId>;
+        StorageMap<_, Blake2_128Concat, T::SmartContract, DAppInfo<T::AccountId>>;
 
-    /// Total block rewards for the pallet per era and total staked funds
+    /// Legacy, don't use.
```
akru, line 171: Please add `TODO` to remove it in the future.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -178,7 +188,25 @@ pub mod pallet {
...
         Twox64Concat,
         EraIndex,
-        EraStakingPoints<T::AccountId, BalanceOf<T>>,
+        EraStakingPoints<BalanceOf<T>>,
+    >;
+
+    /// Era in which dapp reward was last claimed.
+    #[pallet::storage]
+    #[pallet::getter(fn last_claimed_era)]
+    pub type LastClaimedEra<T: Config> =
```
akru, line 197: So, this is for dapps to do sequential claiming, right?

akru (changes requested):

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Yes, I created this PR just to showcase how we could modify `claim_dapp` to avoid specifying era.
It shows lots of changes now because base branch changed...

I didn't want to push these changes into the original commit:
https://github.com/AstarNetwork/astar-frame/pull/6

Sorry for the confusion! I see that you've left comments here but they can be applied to the original commit, no problem.

frame/dapps-staking/src/lib.rs
Dinonard: I agree and don't like this name at all.
It's remnant from the legacy.

Personally, I'd like to get rid of *Points* entirely since there are none in the struct.

Perhaps `DappStakingInfo` or `ContractStakingInfo`?

frame/dapps-staking/src/lib.rs
Dinonard: I expanded all of the comments for the new structs.

I should have done it before, sorry. My focus was on explaining how it works in the notion document.

frame/dapps-staking/src/lib.rs
Dinonard: I provided more information and examples. Hope they make it more clear now!

frame/dapps-staking/src/lib.rs
Dinonard: You're right, that's way better!

frame/dapps-staking/src/lib.rs
Dinonard: Done

frame/dapps-staking/src/lib.rs
Dinonard: Done

frame/dapps-staking/src/lib.rs
Dinonard: We can change it to `RewardDestination`, sure!

For options, I believe `Stake` doesn't tell the entire story since we'll also deposit rewards into `Balance`.
I'd like to avoid people thinking that if they _re-stake_, the reward doesn't go immediately to their balance.

Would `Balance` and `StakeAndBalance` work?

Dinonard (commented):
Comments have been addressed [here](https://github.com/AstarNetwork/astar-frame/pull/6/files/30ebaedb28d1411460f0cddec330d1d86d9d9741..aab4c33324cffb1fc553ca7f01a4ed25a8af82d4).

I've singled out the two commits which only contain changes related to your comments so it's easier to review it again.

I will modify the names for reward destination and staking points once we agree on them.

frame/dapps-staking/src/lib.rs
akru: So, may be we could drop `Staking` from the name, whats about `DappInfo`?

frame/dapps-staking/src/lib.rs
akru: May be `StakeBalance` and `FreeBalance` in that case?

frame/dapps-staking/src/lib.rs
Dinonard: We already have that one [here](https://github.com/AstarNetwork/astar-frame/pull/17/files/b06c0c514da703841618eea521477aee3d1a2ae7#diff-74f8acb4bf207fcaa637a96dcba0e2d302253ac759392b2799bf2f5a32938745R52-R58).

frame/dapps-staking/src/lib.rs
Dinonard: Sure, sounds good.

frame/dapps-staking/src/lib.rs
Dinonard: Any other idea maybe?

I'm for `ContractStakeInfo`.

Dinonard:
Deprecated.

PR #18, opened 2022-02-23
Concept - Dynamic Inflation Storage Schema for DApp Staking

Dinonard:
**Pull Request Summary**

The idea behind this PR is to show how storage would be modified to accommodate dynamic inflation.
The key concept is that we have separate storage for:
1.  stakers reward
2.  dapps reward

This allows us to control two separate reward pools from the runtime code.

E.g. as TVL changes,


Dinonard:
Storage schema has been integrated into individual claim PR.

PR #21, opened 2022-02-23
Preparation for staking festival end on Astar

Dinonard:
**Pull Request Summary**

This PR isn't a kill switch :)
The purpose is to provide call to disable pallet - this means that users won't be able to stake or claim rewards anymore.
Therefore, it is important that all rewards have been claimed prior to putting the pallet into maintenance mode.

The flow for ending the staking festival is as follows:
* uprgrade Astar with new runtime (maintenance mode support)
* claim all pending rewards
* enable maintenance mode for the pallet
* unregister all dapps (this will unlock the entire TVL)
* purge all dapps-staking pallet related DB entries
* upgrade Astar with new runtime that contains individual claim

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


akru:
Generally, maintenance mode looks good, but it’s not clear how it will works for ending festival. And in that cases we will use this kill switch. Could you clear motivation please.


As general solution for temporary disabling pallet methods (like register or stake) we could use internal substrate call filters: https://github.com/AstarNetwork/Astar/blob/master/runtime/astar/src/lib.rs#L152

Dinonard:
@akru sure, that's a good point. My idea was the following:

1. ensure all rewards are claimed (we can manually check each contract for this)
2. enter maintenance mode (era changes and extrinsic calls no longer work)
3. unregister all contracts using the `unregister` call which will remain operational in maintenance mode
4. ensure that no more `Ledger` entries exist (meaning no locks on funds remain)
5. use `pallet_system::kill_prefix` to remove all entries related to dapps-staking

I was also thinking of introducing `maintenance_mode` concept into _latest_ pallet version.
It could be very beneficial to be able to _turn off_ pallet in case we detect some error.
What do you think about that?


For staking-festival, we can use the approach you linked. I didn't know about that so far.
I guess we could disable everything except for `unregister` call.
I'd also need to modify some `on_initialize` and `on_nonzero_unbalanced`.

Dinonard:
I've updated the PR description in order to make it clear what the idea behind PR is.
Please comment and let's upgrade Astar soon with new functionality.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -131,6 +131,9 @@ pub mod pallet {
     #[pallet::getter(fn force_era)]
     pub type ForceEra<T> = StorageValue<_, Forcing, ValueQuery, ForceEraOnEmpty>;
 
+    #[pallet::storage]
+    pub type PalletDisabled<T> = StorageValue<_, ()>;
```
hoonsubin: Noob question, but is there a reason that you use a pair of some and none instead of a boolean value?

hoonsubin: Also, how will this look on the client side?

Dinonard: Changed to `bool` now.

Dinonard:
To make updates of Astar dapps-staking pallet more _future-friendly_, I suggest we just use this branch `legacy/dapps-staking-astar` going forwad.

So after this PR is reviewed/approved, I will just close it and continue using the branch.

Dinonard:
We can end the festival without the use of this PR.

PR #23, opened 2022-03-08
No-loop dapps-staking

Dinonard:
**Pull Request Summary**

Removed _hidden_ iteration via for-loop from dapps-staking.
`iter_key_prefix()` reads in all keys which makes its usage very expensive and uncontrollable in weight and PoV size.

The idea behind this commit is to show a most simple solution on how this problem can be solved.

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -189,7 +189,7 @@ pub mod pallet {
 
     /// Stores amount staked and stakers for a contract per era
     #[pallet::storage]
-    #[pallet::getter(fn contract_era_stake)]
+    #[pallet::getter(fn contract_stake_info)]
     pub type ContractEraStake<T: Config> = StorageDoubleMap<
         _,
         Blake2_128Concat,
@@ -318,7 +318,7 @@ pub mod pallet {
             // a new era is triggered. This code is just a safety net to ensure nothing is broken
             // if we fail to do that.
             if PalletDisabled::<T>::get() {
-                return Zero::zero();
+                return T::DbWeight::get().reads(1);
             }
 
             let force_new_era = Self::force_era().eq(&Forcing::ForceNew);
@@ -335,15 +335,18 @@ pub mod pallet {
 
                 let reward = BlockRewardAccumulator::<T>::take();
                 Self::reward_balance_snapshoot(previous_era, reward);
+                let consumed_weight = Self::rotate_staking_info(previous_era);
 
                 if force_new_era {
                     ForceEra::<T>::put(Forcing::NotForcing);
                 }
 
                 Self::deposit_event(Event::<T>::NewDappStakingEra(next_era));
-            }
 
-            T::DbWeight::get().writes(5)
+                consumed_weight + T::DbWeight::get().reads_writes(5, 3)
+            } else {
+                T::DbWeight::get().reads(4)
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -965,34 +968,45 @@ pub mod pallet {
...
+                } else {
+                    consumed_weight = consumed_weight.saturating_add(T::DbWeight::get().reads(1));
+                }
+            }
+
+            consumed_weight
+        }
+
+        /// This helper returns `ContractStakeInfo` for given era or default value if storage have no data for it.
         pub fn staking_info(
```
akru: May be we can drop this helper because it become just an alias for storage read call.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -965,34 +968,45 @@ pub mod pallet {
...
+        /// Used to copy all `ContractStakeInfo` from the ending era over to the next era.
+        /// This is the most primitive solution since it scales with number of dApps.
+        /// It is possible to provide a hybrid solution which allows laziness but also prevents
+        /// a situation where we don't have access to the required data.
+        fn advance_staking_info(current_era: EraIndex) -> u64 {
+            let next_era = current_era + 1;
+
+            let mut consumed_weight = 0;
+
+            for (contract_id, dapp_info) in RegisteredDapps::<T>::iter() {
```
akru, line 983: Whats about to iterate over `ContractEraStake` for current era? It will save some weight.

akru: So, the cleanup unregistered dapps could be an issue in that case.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -965,34 +968,45 @@ pub mod pallet {
             GeneralEraInfo::<T>::insert(era, era_info);
         }
 
-        /// This helper returns `ContractStakeInfo` for given era if possible or latest stored data
-        /// or finally default value if storage have no data for it.
+        /// Used to copy all `ContractStakeInfo` from the ending era over to the next era.
+        /// This is the most primitive solution since it scales with number of dApps.
+        /// It is possible to provide a hybrid solution which allows laziness but also prevents
+        /// a situation where we don't have access to the required data.
+        fn advance_staking_info(current_era: EraIndex) -> u64 {
```
akru: May be `rotate_staking_info`?

frame/dapps-staking/src/pallet/mod.rs
akru: But unregistered dapps should have staking info for user unstake, didn't? We can cleanup data when stake becomes `0`.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Sure, I kept it just in order to keep changes for this PR version to the minimum.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: The key format is: <contract_id, era> so we cannot iterate over prefix, unfortunately.

We cannot clean data when stake becomes `0` since we need to know when it became `0`.
It's important we keep it in storage (up to history depth).

frame/dapps-staking/src/pallet/mod.rs
Dinonard: I can rename it to that, sure, but we don't just _rotate_, we also do copy.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Oh, I misunderstood - you meant `unregistered` dapps.

I was thinking of handling this in a bit different way as a part of future improvement - add a cleanup extrinsic call that will remove all data related to a particular unregistered dapp after some time.
Otherwise, as you noted, we would keep data for unregistered dapp forever.

Maar-io (approved):
lgtm

PR #27, opened 2022-03-15
Dynamic Inflation

Dinonard:
**Pull Request Summary**

Introduces a new pallet, `pallet-reward-distribution`, to provide a dynamic way of distributing block rewards.
The idea is to have reward distribution parametrized and configurable via root-only call.
This is different from current situation since if we want to modify anything, we need to do an runtime upgrade.

**TODO**
- [x] initial implementation (for getting feedback) 
- [x] benchmarks
- [x] documentation
- [x] unit tests
- [x] initial reward config genesis?
- [x] test with real runtime
- [x] adapt dapps-staking public interface
- [x] merge (or remove?) block-reward pallet

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [x] documentation changes


frame/reward-distribution/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// TODO (remove this)
+// Alternative to this approach would be to have some as configurable constants.
+// This approach, however, provides more flexibility since we can change them without
+// runtime upgrade.
+
+/// List of parameters used to calculate reward distribution portions for all the beneficiaries.
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub struct RewardDistributionParams {
```
akru: Looks like RewardDistributionConfig is more suitable name for this struct

frame/reward-distribution/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+            collators_percent: Perbill::from_percent(10),
+            adjustable_percent: Zero::zero(),
+            ideal_dapps_staking_tvl: Zero::zero(),
+        }
+    }
+}
+
+impl RewardDistributionParams {
+    /// `true` if sum of all percentages is `one whole`, `false` otherwise.
+    fn verify(&self) -> bool {
```
akru: May be `is_consistent` a bit more descriptive rather than `verify`.

frame/reward-distribution/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+            .base_treasury_percent
+            .saturating_add(self.base_staker_percent)
+            .saturating_add(self.dapps_percent)
+            .saturating_add(self.collators_percent);
+
+        (Perbill::one() - sum) == self.adjustable_percent
+    }
+}
+
+pub trait TvlProvider<Balance> {
```
akru: Please add trait description that explains the cause of introducing this trait.

frame/reward-distribution/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Total value locked as amount
+    fn tvl() -> Balance;
+
+    ///Total value locked as percentage of total issuance
+    fn tvl_percentage() -> Perbill;
```
akru: Why don't use `Currency` trait as base?

akru:
In general I feel that this pallet should replace `block-reward` pallet, we don't need to have many pallet that handle rewards, it's difficult to track reward path. So, instead of distribution on runtime code scope let's delegate distribution to dedicated pallet with fine tuned config.

One suggestion, instead of beneficiary trait why don't use pallet config types for reward beneficiaries, like `type DappStakingDevelopers: OnImbalance<Balance>;`

frame/reward-distribution/src/lib.rs
akru: Or as parameter may be, something like `TvlProvider<Currency>`. In that case `tvl_percentage` could be defined by default.

frame/reward-distribution/src/lib.rs
Dinonard: Sure.

frame/reward-distribution/src/lib.rs
Dinonard: Changed.

frame/reward-distribution/src/lib.rs
Dinonard: Sure, I haven't really documented the pallet in details since this was just supposed to be a PR to give an overview of the idea.

In the end, pallet will have proper documentation everywhere.

frame/reward-distribution/src/lib.rs
Dinonard: I can use `Currency` as base trait but not sure I understand what would be the benefit over current solution.

Regarding `tvl_percentage`, I think we can remove this function since this pallet already has `Currency` as associated type so we can fetch `total_issuance()`.

I will push a new commit now and change this. Please tell me if you had something else in mind since I think I might be misunderstanding the comment.

Dinonard:
> In general I feel that this pallet should replace `block-reward` pallet, we don't need to have many pallet that handle rewards, it's difficult to track reward path. So, instead of distribution on runtime code scope let's delegate distribution to dedicated pallet with fine tuned config.
> 
> One suggestion, instead of beneficiary trait why don't use pallet config types for reward beneficiaries, like `type DappStakingDevelopers: OnImbalance<Balance>;`

That sounds ok - I was thinking the same thing but wanted to avoid doing it initially in order to keep the change smaller.
Let's have a single pallet for handling block-reward then. :+1: 

For the `Beneficiaries` trait, my main motivation was to avoid having 3 small structs which implement `OnImbalanced` trait in the runtimes - instead I opted for a single trait which would enforce a single struct to implement all required functions.
Do you think it's better to have 3 than 1? I don't really mind tbh, just thought it would look cleaner with 1.

Maar-io (approved):
LGTM

frame/reward-distribution/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    impl<T: Config> OnUnbalanced<NegativeImbalanceOf<T>> for Pallet<T> {
+        /// Distribute reward between beneficiaries.
+        ///
+        /// # Arguments
+        /// * `reward` - reward that will be split and distributed
+        ///
+        /// # Returns
+        /// * `weight` of the operation
+        ///
```
Dinonard: Remove this old comment

frame/reward-distribution/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    impl<Moment, T: Config> OnTimestampSet<Moment> for Pallet<T> {
+        fn on_timestamp_set(_moment: Moment) {
+            let inflation = T::Currency::issue(T::RewardAmount::get());
+            Self::on_nonzero_unbalanced(inflation);
+        }
+    }
+
+    impl<T: Config> OnUnbalanced<NegativeImbalanceOf<T>> for Pallet<T> {
```
Dinonard: Perhaps `OnUnbalanced` usage should be dropped since I don't want to expose this to the outside - pallet is responsible both for issuing and distributing rewards.

PR #33, opened 2022-03-29
Staking festival end storage cleanup

Dinonard:
**Pull Request Summary**

Support for clearing old storage items from Astar once we end the staking festival.
Solution has been tested on local network and works fine.

### Upgrade Flow
1. Using `sudo`, unregister all contracts
2. Use binary built with `try-runtime` features to simulate runtime upgrade logic on Astar and confirm it works
3. Initiate runtime upgrade for Astar network with latest dapps-staking
4. Once new runtime is applied, pallet will enter **maintenance mode**
5. We will need to call `do_upgrade` call a few times in order to cleanup old storage items
6. Once last item is cleared, **maintenance mode** will be disabled an pallet will resume normal operation
7. :tada: :tada: :tada: we've re-launched!

**TODO**
- [x] beef up era 1 reward
- [x] decide unbonding period for Astar

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/frame/dapps-staking/src/migrations.rs
+++ b/frame/dapps-staking/src/migrations.rs
@@ -1077,3 +1077,234 @@ pub mod v3 {
         Ok(())
     }
 }
+
+pub mod festival_end {
+
+    use super::*;
+    use frame_support::{log, storage::child::KillStorageResult, traits::Get, weights::Weight};
+    use sp_runtime::traits::Zero;
+
+    #[cfg(feature = "try-runtime")]
+    pub fn pre_migrate<T: Config>() -> Result<(), &'static str> {
+        assert!(Ledger::<T>::iter().count().is_zero());
+        assert!(BlockRewardAccumulator::<T>::exists());
+
+        // Sanity check for pallet free balance. UT covers this but not via live chain check.
+        let halved_unit: BalanceOf<T> = 1_000_000_000_u32.into();
+        let unit = halved_unit * halved_unit;
+        let threshold_balance = unit * 25_000_000_u32.into();
+
+        assert!(T::Currency::free_balance(&Pallet::<T>::account_id()) > threshold_balance);
+
+        Ok(())
+    }
+
+    /// Should be executed during runtime upgrade
+    /// Won't do the entire cleanup, instead it will just do necessary preparatory actions.
+    pub fn on_runtime_upgrade<T: Config>(weight_limit: Weight) -> Weight {
+        if Ledger::<T>::iter_keys().peekable().peek().is_some() {
+            log::warn!(">>> There are existing account ledgers. This upgrade is invalid!");
+            return T::DbWeight::get().reads(1);
+        }
+
+        // 1. Accumulated rewards for the ongoing era are killed - however, funds will remain in pallet account
+        BlockRewardAccumulator::<T>::kill();
+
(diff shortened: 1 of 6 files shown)


frame/dapps-staking/src/migrations.rs
```diff
@@ -1077,3 +1077,169 @@ pub mod v3 {
...
+        assert!(Ledger::<T>::iter().count().is_zero());
+        assert!(BlockRewardAccumulator::<T>::exists());
+
+        Ok(())
+    }
+
+    /// Should be executed during runtime upgrade
+    /// Won't do the entire cleanup, instead it will just do necessary preparatory actions.
+    pub fn on_runtime_upgrade<T: Config>(weight_limit: Weight) -> Weight {
+        for _ in Ledger::<T>::iter() {
```
codespool: just curious - this method of checking is `Ledger` empty is used because there's no `is_empty` or similar trait on StorageMap, and is more efficient that `count()` (since it returns after first item is found)? Is there some other reason?

frame/dapps-staking/src/migrations.rs
```diff
@@ -1077,3 +1077,169 @@ pub mod v3 {
...
+            log::warn!(">>> There are existing account ledgers. This upgrade is invalid!");
+            return T::DbWeight::get().reads(1);
+        }
+
+        // 1. this needs to be handled immediately since we want to accumulate rewards in the follow-up blocks
```
codespool: we want or don't want? (we are killing the accumulator, so I assume we don't)

frame/dapps-staking/src/migrations.rs
Dinonard: Yes, I did it for the reasons you mentioned :)

Since iterator is lazy, this seems to be best way to check from what I've found.

frame/dapps-staking/src/migrations.rs
Dinonard: Leftover comment from previous iteration - there's actually nothing to save since we're killing the ongoing era and starting from zero.

I will modify it.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -338,6 +338,28 @@ pub mod pallet {
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
+        #[pallet::weight(T::BlockWeights::get().max_block / 5 * 3)]
+        pub fn do_upgrade(
+            origin: OriginFor<T>,
+            weight_limit: Option<Weight>,
+        ) -> DispatchResultWithPostInfo {
+            ensure_signed(origin)?;
```
codespool, line 346: shouldn't this be done by root or sudo?

Dinonard: No, we set it intentionally so anyone can do it.
This makes it easier to do calls since e.g. I can do it without sudo access.

They way call is written, it's very expensive so it cannot be spammed.
Also, there are safeguards in place which basically make the call obsolete once migration is finished.

codespool: ah, got it, thanks for the explanation 👍

codespool (approved):
LGTM

akru (approved):
LGTM

PR #41, opened 2022-04-13
Nomination Transfer

Dinonard:
# Nomination Transfer

## Summary

A new functionality is introduced for the dapps-staking pallet - `nomination transfer`.
The idea is to allow stakers to transfer their nomination from `contract_A` to `contract_B` while bypassing the unbonding period.
This is great for stakers who have already staked on some contract but found another one they would like to nominate instead. Prior to  `nomination_transfer`, they would have to wait first for the unbonding period to complete, not earning any staking rewards, before being able to re-stake their rewards.

## Requirements (useful for UI)

There are certain requirements that must be met in order for nomination transfer to work:
- there must be available nomination transfer charges
  - this requires storage read and removal of old cooldown blocks before checking
- both contracts must be in `Registered` state
- origin contract must differ from the target contract
- rest of the requirements are same as for `bond_and_stake` and `unbond_and_unstake`
  - only difference is that there's no check for `unbonding_period` since it's bypassed here
  - it is possible that if staker isn't already staking on target contract, there are no staking slots left
  - same goes for the amount being transferred - it might be insufficient if this is first stake on the target contract
  - `EraStake` values will be produced both for origin and target contract and will be checked against `MaxEraStakeValues`
  - if nomination transfer amount will take staker below minimum staking amount on the origin contract, staker will be fully unstaked from the origin contract and entire amount will be transferred to the target contract

**TODO**
- [x] run benchmarks & update weights in the pallet
- [x] update precompiles interface & test
- [x] prepare demo video for the team
- [x] discuss how to handle cooldowns (REMOVED)
- [x] create a backlog item for re-staking unbonding chunks?
- [ ] re-run benchmarks prior to merge

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -93,27 +93,6 @@ pub mod pallet {
         #[pallet::constant]
         type MaxEraStakeValues: Get<u32>;
 
-        /// Number of blocks that need to pass after nomination transfer charge has been used before it can be used again.
-        #[pallet::constant]
-        type NominationTransferCooldown: Get<BlockNumberFor<Self>>;
-
-        /// Number of nomination transfer charges a staker can use. Using a charge triggers a cooldown period for it.
-        ///
-        /// # Example
-        /// ```nocompile
-        /// NominationTransferCharges = 2
-        ///
-        /// // One nomination transfer charge has been spent
-        /// nomination_transfer(alice_account, contract_A, contract_B, 100);
-        ///
-        /// // Second nomination transfer charge has been spent
-        /// nomination_transfer(alice_account, contract_A, contract_C, 30);
-        ///
-        /// // Alice needs to wait for the cooldown period until she can use nomination transfer again
-        /// ```
-        #[pallet::constant]
-        type NominationTransferCharges: Get<u32>;
-
         /// The overarching event type.
         type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 
@@ -219,17 +198,6 @@ pub mod pallet {
     pub(crate) type PreApprovedDevelopers<T: Config> =
         StorageMap<_, Twox64Concat, T::AccountId, (), ValueQuery>;
 
-    /// Nomination transfer charges cooldowns ending blocks.
-    #[pallet::storage]
-    #[pallet::getter(fn nomination_transfer_cooldowns)]
-    pub(crate) type NominationTransferCooldowns<T: Config> = StorageMap<
-        _,
-        Blake2_128Concat,
-        T::AccountId,
-        BoundedVec<BlockNumberFor<T>, T::NominationTransferCharges>,
-        ValueQuery,
-    >;
-
     #[pallet::event]
     #[pallet::generate_deposit(pub(crate) fn deposit_event)]
     pub enum Event<T: Config> {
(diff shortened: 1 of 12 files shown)


frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -198,6 +219,17 @@ pub mod pallet {
     pub(crate) type PreApprovedDevelopers<T: Config> =
         StorageMap<_, Twox64Concat, T::AccountId, (), ValueQuery>;
 
+    /// Nomination transfer charges cooldowns ending blocks.
```
akru: Please elaborate why this delay really required, is balance migrations have some uncalculated weights?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -659,6 +665,90 @@ pub mod pallet {
...
+        ///
+        /// Same rules as for `bond_and_stake` and `unbond_and_unstake` apply.
+        /// Minor difference is that there is no unbonding period so this call won't
+        /// check whether max number of unbonding chunks is exceeded.
+        ///
+        #[pallet::weight(T::WeightInfo::nomination_transfer())]
+        pub fn nomination_transfer(
+            origin: OriginFor<T>,
+            origin_contract_id: T::SmartContract,
+            #[pallet::compact] value: BalanceOf<T>,
```
akru, line 643: Let’s permit all balance transfers only, it will make code much more simple and safe.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Discussed in slack - once we agree how to proceed, I will update (probably remove) this part.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: We can do that, it's no problem to implement it.
However, we won't reduce complexity of the code.
We will only reduce what the user can do and possibly I'll have to redo UTs later if users request this feature.

The `bond_and_stake` and `unbond_and_unstake` logic is completely reused here.
The only _new_ part of the code are the checks specific for the nomination transfer.

So if we decide only to allow full transfer, this extrinsic call will change to this:
```Rust
pub fn nomination_transfer(
            origin: OriginFor<T>,
            origin_contract_id: T::SmartContract,
            target_contract_id: T::SmartContract,
        // same code as before...
        let mut origin_staker_info = Self::staker_info(&staker, &origin_contract_id);
        let value = origin_staker_info.latest_staked_value();
        // same code as before
}
```

Dinonard: As discussed on dev sync, we'll keep the configurable value.

codespool (commented):
LGTM 👍

codespool (approved):

Maar-io:
My only comment is usability of the `precompile` changes to support `withdraw_from_unregistered` and `nomination_transfer`. Maybe Astar portal will have use of these but I do not see why would evm smart contract use these calls.

Maar-io (approved):
LGTM

Dinonard:
@Maar-io for the `withdraw_from_unregistered`, if contract is staking on another contract that gets unregistered, they cannot get their staked tokens out otherwise.

For `nomination_transfer`, I added it for consistency.

PR #43, opened 2022-04-22
Reserve based slashing for collator selection pallet

akru:
**Pull Request Summary**

> Describe what changes this pull request makes to the repository

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)

### DIFF
--- a/frame/collator-selection/src/lib.rs
+++ b/frame/collator-selection/src/lib.rs
@@ -26,12 +26,12 @@
 //! ## Terminology
 //!
 //! - Collator: A parachain block producer.
-//! - Bond: An amount of `Balance` _locked_ for candidate registration.
+//! - Bond: An amount of `Balance` _reserved_ for candidate registration.
 //! - Invulnerable: An account guaranteed to be in the collator set.
 //!
 //! ## Implementation
 //!
-//! The final [`Collators`] are aggregated from two individual lists:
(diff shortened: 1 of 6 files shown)


bLd75 (approved):

frame/collator-selection/src/lib.rs
```diff
@@ -255,6 +263,7 @@ pub mod pallet {
         NewCandidacyBond(BalanceOf<T>),
         CandidateAdded(T::AccountId, BalanceOf<T>),
         CandidateRemoved(T::AccountId),
+        CandidateSlashed(T::AccountId),
```
Dinonard, line 266: Would it be relevant to emit the slashed amount?

Dinonard (approved):

frame/collator-selection/src/lib.rs
akru: It's constant, not sure that it's required

Dinonard: Oh right, you get kicked out afterwards.

I was thinking it was possible for it to happen multiple times before collator is removed.

PR #44, opened 2022-04-25
Compound reward precompile

Dinonard:
**Pull Request Summary**

Adds precompile support for changing reward destination for the compounding reward PR.

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/precompiles/dapps-staking/src/lib.rs
+++ b/precompiles/dapps-staking/src/lib.rs
@@ -10,6 +10,7 @@ use frame_support::{
     dispatch::{Dispatchable, GetDispatchInfo, PostDispatchInfo},
     traits::{Currency, Get},
 };
+use pallet_dapps_staking::RewardDestination;
 use pallet_evm::{AddressMapping, Precompile};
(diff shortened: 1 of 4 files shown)


precompiles/dapps-staking/src/lib.rs
```diff
@@ -453,6 +480,18 @@ where
...
+                input,
+                gasometer,
+                context,
+            )?,
+            Action::StakeBalanceRewardDestination => Self::set_reward_destination(
+                RewardDestination::StakeBalance,
+                input,
+                gasometer,
+                context,
+            )?,
```
Maar-io: Why not use input parameter instead of 2 methods?

Dinonard: I will change it but Solidity enums are primitive (represented by u8) so if we introduce more complex enum type in the future, we'll need to add a different method.

Maar-io (commented):
LGTM

PR #48, opened 2022-05-02
Dapps-staking precompile: read staked amount on the contract

Maar-io:
**Pull Request Summary**

> This functionality is an improvement. It adds possibility to read staked amount on the contract for the user (staker).
> Until now it was only possible to read if user (staker) is staking in general terms.

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)

### DIFF
--- a/precompiles/dapps-staking/DappsStaking.sol
+++ b/precompiles/dapps-staking/DappsStaking.sol
@@ -29,6 +29,12 @@ interface DappsStaking {
     /// @return amount, Staked amount by the staker
     function read_staked_amount(bytes calldata staker) external view returns (uint128);
 
+    /// @notice Read Staked amount on a given contract for the staker
+    /// @param contract_id contract evm address
+    /// @param staker in form of 20 or 32 hex bytes
+    /// @return amount, Staked amount by the staker
+    function read_staked_amount_on_contract(address contract_id, bytes calldata staker) external view returns (uint128);
+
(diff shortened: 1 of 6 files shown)


precompiles/dapps-staking/DappsStaking.sol
```diff
@@ -29,6 +29,12 @@ interface DappsStaking {
     /// @return amount, Staked amount by the staker
     function read_staked_amount(bytes calldata staker) external view returns (uint128);
 
+    /// @notice Read Staked amount on a given contract for the staker
+    /// @param contract_id contract evm address
+    /// @param staker in form of 20 or 32 hex bytes
+    /// @return amount, Staked amount by the staker
+    function read_staked_amount_on_contract(address contract_id, bytes calldata staker) external view returns (uint128);
```
Dinonard, line 36: This won't be WASM compatible.

But it also shouldn't be a problem since no EVM contract can query WASM related stuff yet since there's nothing to really query.

Dinonard (approved):
Only possibly problematic thing is compatibility with future contract types (e.g. WASM).

Also, make sure to merge this into `polkadot-v0.9.19` too.
Most likely that will be the branch from which we'll build the new runtime upgrades.

PR #51, opened 2022-05-06
Fix for incorrect era when compounding

Dinonard:
**Pull Request Summary**

Reproduced and fixed issue where incorrect contract era stake is overwritten.

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/frame/dapps-staking/src/fix.rs
+++ b/frame/dapps-staking/src/fix.rs
@@ -17,6 +17,22 @@ pub mod restake_fix {
     use sp_runtime::traits::{Saturating, Zero};
     use sp_std::collections::btree_map::BTreeMap;
 
+    // Temp struct of corrected ContractStakeInfo records with progress data
+    #[derive(Clone, PartialEq, Encode, Decode, RuntimeDebug, TypeInfo, Default)]
+    pub struct RestakeFix<Balance: HasCompact> {
+        // store progress so iteration can continue in the next block
+        last_processed_staker: Option<Vec<u8>>,
+        contract_staking_info: BTreeMap<Vec<u8>, ContractStakeInfo<Balance>>,
+        // should be flipped once the process is complete
+        all_stakers_processed: bool,
+    }
+
+    pub fn restake_fix_migration<T: Config>(weight_limit: Weight) -> Weight {
+        let mut restake_fix = RestakeFixAccumulator::<T>::get();
+
+        T::DbWeight::get().reads(1)
+    }
+
     #[cfg(feature = "try-runtime")]
     pub fn post_migrate<T: Config>() -> Result<(), &'static str> {
         // Pallet should be enabled after we finish
--- a/frame/dapps-staking/src/fix.rs
+++ b/frame/dapps-staking/src/fix.rs
@@ -30,6 +30,25 @@ pub mod restake_fix {
     pub fn restake_fix_migration<T: Config>(weight_limit: Weight) -> Weight {
         let mut restake_fix = RestakeFixAccumulator::<T>::get();
 
+        // read all_stakers_processed
+
+        // if false:
+
+        // read ledger from last_processed_staker or first if None
+
+        // for each record add to contract_staking info (amount and count)
+        // and add read weight
+
+        // when weight hits limit, write and return
+
+        // if true
+
+        // if contractStakeInfo is empty, we're done
+
+        // for each ContractStakeInfo in RestakeFix
+        // write to ContractEraStake until weight hits limit
+        // delete used records
+
         T::DbWeight::get().reads(1)
     }
 
--- a/frame/dapps-staking/src/fix.rs
+++ b/frame/dapps-staking/src/fix.rs
@@ -1,20 +1,13 @@
 use super::*;
-use codec::{Decode, FullCodec};
-use frame_support::storage::unhashed;
-use pallet::pallet::*;
-use sp_std::fmt::Debug;
+use codec::Decode;
 
 pub mod restake_fix {
 
     use super::*;
     use codec::Encode;
     use frame_support::log;
-    use frame_support::{
-        storage::generator::{StorageDoubleMap, StorageMap},
-        traits::Get,
-        weights::Weight,
-    };
-    use sp_runtime::traits::{Saturating, Zero};
+    use frame_support::{storage::generator::StorageMap, traits::Get, weights::Weight};
+    use sp_runtime::traits::Saturating;
     use sp_std::collections::btree_map::BTreeMap;
 
     // Temp struct of corrected ContractStakeInfo records with progress data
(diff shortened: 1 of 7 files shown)


frame/dapps-staking/src/testing_utils.rs
```diff
@@ -478,38 +477,62 @@ pub(crate) fn assert_claim_staker(claimer: AccountId, contract_id: &MockSmartCon
...
-            init_state.era_info.staked + reward,
-            final_state.era_info.staked
+            init_state_current_era.era_info.locked + reward,
+            final_state_current_era.era_info.locked
         );
         assert_eq!(
-            init_state.era_info.locked + reward,
-            final_state.era_info.locked
+            init_state_current_era.contract_info.total + reward,
+            final_state_current_era.contract_info.total
```
Dinonard, line 516: This check was missing.

codespool (approved):

codespool:
@Dinonard I got the logic in, but I'm having trouble with the `contract` type and converting it to `Vec<u8>`, mainly because I couldn't find a method to get it in raw format from `iter_prefix_from`. (if I used the `storage_map_final_key` from ledger like you suggested, it complained there are encodeLike traits missing on `SmartContract` type, and if I added that trait, then the runtime build failed - i figured we don't want to touch runtime code for this)

I tried to use `SmartContract` type as key for BTreeMap so I don't have to convert, but that got me an overflow issue from the scale codec.

In the end I used I used `PrefixIterator::last_raw_key` but that got me into ownership issues.

I know there's probably a super simple solution, but I spent a good part of last night trying to find it, so any advice is welcome 🙏

frame/dapps-staking/src/fix.rs
```diff
@@ -0,0 +1,175 @@
...
+        // accumulate ContractStakeInfo by contract address
+        contract_staking_info: BTreeMap<Vec<u8>, ContractStakeInfo<Balance>>,
+        // flip to true when all Ledger data has been processed
+        all_stakers_processed: bool,
+    }
+
+    pub fn restake_fix_migration<T: Config>(weight_limit: Weight) -> Weight {
+        log::info!("Executing a step of restake fix migration.");
+        let mut restake_fix = RestakeFixAccumulator::<T>::get();
+        // one read for RestakeFixAccumulator
```
Dinonard, line 31: I have to comment this part because it made the review quite difficult - most of the comments you have in this function are redundant, imo. The code is enough self-explanatory.

I checked it out locally and removed all comments in order to make it easier for myself.
Of course, it might just be me.

frame/dapps-staking/src/fix.rs
```diff
@@ -0,0 +1,175 @@
...
+                // every iteration adds one more read
+                consumed_weight = consumed_weight.saturating_add(T::DbWeight::get().reads(1));
+                // take iter for inner loop from start or last processed contract
+                let contract_iter = if let Some(last_processed_contract) =
+                    restake_fix.last_processed_staker_contract.clone()
+                {
+                    GeneralStakerInfo::<T>::iter_prefix_from(&staker, last_processed_contract)
+                } else {
+                    GeneralStakerInfo::<T>::iter_prefix(&staker)
+                };
```
Dinonard, line 55: You can remove this and process the entire batch for each staker - even if you go over the weight limit, it will be negligible.

This will also simplify your logic a lot.

frame/dapps-staking/src/fix.rs
```diff
@@ -0,0 +1,175 @@
...
+                for (contract, staking_info) in contract_iter {
+                    log::info!(
+                        "Extracting contract stake data for staker address: {:?}.",
+                        staker
+                    );
+                    // every inner iteration also adds one read
+                    consumed_weight = consumed_weight.saturating_add(T::DbWeight::get().reads(1));
+                    let staked_value = staking_info.latest_staked_value();
```
Dinonard, line 63: What if latest staked value is zero?

frame/dapps-staking/src/fix.rs
```diff
@@ -0,0 +1,175 @@
...
+                    // get contract address as Vec<u8>
+                    let contract_address = PrefixIterator::last_raw_key(&contract_iter);
```
Dinonard, line 65: This is the root of the problem you wrote about. This isn't a good approach.

Instead, you should use `contract.encode()`. Later, you get use `T::SmartContract::decode(&mut &contract_vec[..])` to get back the original value.

frame/dapps-staking/src/fix.rs
```diff
@@ -0,0 +1,175 @@
...
+                        break 'staker_iter;
+                    }
+                }
+
+                // when one staker is processed, store it and clear stored contract
+                restake_fix.last_processed_staker_contract = None;
+                restake_fix.last_fully_processed_staker =
+                    Some(Ledger::<T>::storage_map_final_key(staker));
+            }
+        } else {
```
Dinonard, line 94: As mentioned in the call, this part isn't optimal because there's no reason why we couldn't execute both steps in a single call.

frame/dapps-staking/src/fix.rs
```diff
@@ -0,0 +1,175 @@
...
+                consumed_weight = consumed_weight.saturating_add(T::DbWeight::get().reads(1));
+
+                let staking_info_iter = restake_fix.contract_staking_info.iter();
+
+                for (contract, contract_stake_info) in staking_info_iter {
+                    // write new ContractEraStake from corrected records in contract_staking_info
+                    // and remove that record afterwards
+                    // break out when the weight hits the limit
+                    ContractEraStake::<T>::insert(&contract, current_era, contract_stake_info);
+                    restake_fix.contract_staking_info.remove(contract);
```
Dinonard, line 111: You shouldn't modify map while iterating over it.

Unfortunately, calls like `pop_first` are still experimental so we cannot use them.
And BTreeMap doesn't have a draining iterator.

But what you can do is create a copy of the key set and iterate over it.

frame/dapps-staking/src/fix.rs
```diff
@@ -0,0 +1,175 @@
...
+                            consumed_weight
+                        );
+                        break;
+                    }
+                }
+            }
+        }
+        // store updated accumulator and update total weight for that write
+        RestakeFixAccumulator::<T>::put(restake_fix);
+        consumed_weight = consumed_weight.saturating_add(T::DbWeight::get().writes(1));
```
Dinonard, line 125: Imo, this is incorrect.

If you've reached this point, you should kill `RestakeFixAccumulator` since it should mean you're finished.

You should also disable maintenance mode.

codespool:
@Dinonard thanks for the comments!
Since this is resolved from another PR, I'll just fix them for the sake of learning and close the PR.

akru:
@Dinonard should we close this PR?

PR #52, opened 2022-05-08
Fix/dapps staking restake v2

Dinonard:
**Pull Request Summary**

Fix for issue where compounding rewards implementation would incorrectly override current era info with past era info.
* Added UT to cover the scenarios
* Added new ext call which allows us root-only way to modify `ContractEraStake`

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


Dinonard:
I will merge this one using admin privileges.

Code has been thoroughly tested and we need to move today.

PR #55, opened 2022-05-18
Added pallet-xcm with reserve withdraw function

akru:
**Pull Request Summary**
Technically its fork of Polkadot pallet-xcm with new function to withdraw reserve transfers. 

This PR introduces two methods:
* reserveWithdrawAssets()
* limitedReserveWithdrawAssets()

Both has same arguments as reserveTransfer*** methods, but do reverse action: withdraw asset from reserve, for example, relay chain.

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)

### DIFF
--- a/frame/pallet-xcm/Cargo.toml
+++ b/frame/pallet-xcm/Cargo.toml
@@ -1,0 +1,43 @@
+[package]
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2021"
+name = "pallet-xcm"
+version = "0.9.19"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
+scale-info = { version = "2.1.0", default-features = false, features = ["derive"] }
+serde = { version = "1.0.136", optional = true, features = ["derive"] }
+log = { version = "0.4.16", default-features = false }
+
+sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.19" }
+sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.19" }
+sp-core = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.19" }
+frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.19" }
+frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.19" }
+
+xcm = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.19" }
+xcm-executor = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.19" }
+
+[dev-dependencies]
+pallet-balances = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19" }
+polkadot-runtime-parachains = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.19" }
+sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19" }
+xcm-builder = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.19" }
+polkadot-parachain = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.19" }
+
+[features]
+default = ["std"]
+std = [
+	"codec/std",
+	"scale-info/std",
+	"serde",
+	"sp-std/std",
+	"sp-core/std",
+	"sp-runtime/std",
+	"frame-support/std",
+	"frame-system/std",
+	"xcm/std",
+	"xcm-executor/std",
+]
+runtime-benchmarks = []
--- a/frame/pallet-xcm/Cargo.toml
+++ b/frame/pallet-xcm/Cargo.toml
@@ -40,4 +40,9 @@ std = [
 	"xcm/std",
 	"xcm-executor/std",
 ]
-runtime-benchmarks = []
+runtime-benchmarks = [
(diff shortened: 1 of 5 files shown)


frame/pallet-xcm/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2021"
+name = "pallet-xcm"
+version = "0.9.19"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
+scale-info = { version = "2.0.0", default-features = false, features = ["derive"] }
```
Dinonard: We use `2.1.0` in other pallets. Better to be consistent.

Dinonard (commented):
Could you maybe push 2 commits?

1 base and another that would just be your modifications.
It'd be MUCH easier to review that way.

frame/pallet-xcm/src/lib.rs
```diff
@@ -0,0 +1,1782 @@
...
+					let mut message = Xcm(vec![
+                        WithdrawAsset(assets),
+						InitiateReserveWithdraw { assets: Wild(All), reserve: dest, xcm: Xcm(vec![]) }
+					]);
+					T::Weigher::weight(&mut message).map_or(Weight::max_value(), |w| 100_000_000 + w)
+				},
+				_ => Weight::max_value(),
+			}
+		})]
+        pub fn reserve_withdraw_assets(
```
akru, line 592: @Dinonard Just added this method, no more.

frame/pallet-xcm/Cargo.toml
akru: Fixed, thank you!

frame/pallet-xcm/src/lib.rs
```diff
@@ -0,0 +1,1786 @@
...
+                },
+            ]);
+            let mut message = Xcm(vec![
+                WithdrawAsset(assets),
+                InitiateReserveWithdraw {
+                    assets: Wild(All),
+                    reserve: dest,
+                    xcm,
+                },
+            ]);
```
Dinonard, line 1100: I checked what each line does and it looks good.
I'm just surprised (same as you) that they haven't added support for this.

Will you also make a PR for `substrate`?

frame/pallet-xcm/src/tests.rs
```diff
@@ -0,0 +1,1248 @@
...
+        );
+    });
+}
+
+/// Test `reserve_transfer_assets`
+///
+/// Asserts that the sender's balance is decreased and the beneficiary's balance
+/// is increased. Verifies the correct message is sent and event is emitted.
+#[test]
+fn reserve_transfer_assets_works() {
```
Dinonard, line 425: Please add UT for your `reserve_withdraw_assets`.

Dinonard (changes requested):
Looks good!

I still wish that base code and your changes were separate though.

frame/pallet-xcm/src/lib.rs
akru: Someday lets make PR to polkadot with it.

akru:
@Dinonard reworked

frame/pallet-xcm/src/tests.rs
akru: Added

frame/pallet-xcm/src/tests.rs
```diff
@@ -591,6 +591,59 @@ fn unlimited_reserve_transfer_assets_works() {
...
+        (ParaId::from(PARA_ID).into_account(), INITIAL_BALANCE),
+    ];
+    new_test_ext_with_balances(balances).execute_with(|| {
+        let weight = BaseXcmWeight::get();
+        let dest: MultiLocation = Junction::AccountId32 {
+            network: NetworkId::Any,
+            id: ALICE.into(),
+        }
+        .into();
+        assert_eq!(Balances::total_balance(&ALICE), INITIAL_BALANCE);
```
Dinonard, line 611: Imo, it'd be good to also check the sovereign account initial and final balance.

frame/pallet-xcm/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+	"serde",
+	"sp-std/std",
+	"sp-core/std",
+	"sp-runtime/std",
+	"frame-support/std",
+	"frame-system/std",
+	"xcm/std",
+	"xcm-executor/std",
+]
+runtime-benchmarks = []
```
Dinonard: Please change to:

```
runtime-benchmarks = [
	"xcm/runtime-benchmarks",
	"xcm-builder/runtime-benchmarks",
    "frame-support/runtime-benchmarks",
    "frame-system/runtime-benchmarks",
]
```

This will fix the failing UT compilation.

Dinonard (approved):
LGTM

Please just fix the UT CI.

frame/pallet-xcm/src/lib.rs
Dinonard: Sure, that sounds good.

If we can avoid maintaining this pallet in our repo, it'd be nice.
I'm pretty sure we can expect heavy changes here as new XCM version roll out.

hoonsubin:
Should we also consider adding https://github.com/PureStake/moonbeam/blob/571e842db30fe25d550ae20ea23b3078906a21fc/precompiles/xcm_transactor/XcmTransactor.sol so users can withdraw from EVM?

frame/pallet-xcm/src/lib.rs
hoonsubin: interesting. Maybe they don't want their DOTs back xD

hoonsubin (approved):
LGTM

PR #57, opened 2022-05-30
EVM precompile for Pallet XCM

akru:
**Pull Request Summary**

> Describe what changes this pull request makes to the repository

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)

### DIFF
--- a/precompiles/xcm/Cargo.toml
+++ b/precompiles/xcm/Cargo.toml
@@ -1,0 +1,45 @@
+[package]
+name = "pallet-evm-precompile-xcm"
+authors = [ "Stake Technologies <devops@stake.co.jp>" ]
+description = "Basic XCM support for EVM."
+edition = "2021"
+version = "1.0.0"
+
+[dependencies]
+precompile-utils = { path = "../utils", default-features = false }
+num_enum = { version = "0.5.3", default-features = false }
+log = "0.4.16"
+
+# Substrate
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = [ "max-encoded-len" ] }
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19", default-features = false }
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19", default-features = false }
+sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19", default-features = false }
+
+# Frontier
+fp-evm = { git = "https://github.com/AstarNetwork/frontier", branch = "polkadot-v0.9.19", default-features = false }
+pallet-evm = { git = "https://github.com/AstarNetwork/frontier", branch = "polkadot-v0.9.19", default-features = false }
+
+[dev-dependencies]
+serde = "1.0.100"
+scale-info = "2.0"
+derive_more = "0.99"
+hex-literal = "0.3.4"
+
+frame-support = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19" }
+frame-system = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19" }
+sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19" }
+pallet-balances = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19" }
+pallet-timestamp = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19" }
+
+[features]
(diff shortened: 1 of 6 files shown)


precompiles/xcm/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19", default-features = false }
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19", default-features = false }
+sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.19", default-features = false }
+
+# Frontier
+fp-evm = { git = "https://github.com/AstarNetwork/frontier", branch = "polkadot-v0.9.19", default-features = false }
+pallet-evm = { git = "https://github.com/AstarNetwork/frontier", branch = "polkadot-v0.9.19", default-features = false }
+
+# Polkadot
+xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.19", default-features = false }
```
Dinonard, line 29: Should we use our own fork here?

Not for this PR, but perhaps it would be good to rename our `pallet-xcm` since now it has the exact same name as the one used by polkadot/kusama/etc. runtimes.

precompiles/xcm/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+xcm-executor = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.19", default-features = false }
+
+[dev-dependencies]
+serde = "1.0.100"
+scale-info = "2.0"
```
Dinonard: We use `2.1.0` for other crates.

precompiles/xcm/XCM.sol
```diff
@@ -0,0 +1,19 @@
+pragma solidity ^0.8.0;
+
+/**
+ * @title XCM interface.
+ */
+interface XCM {
+    /**
+     * @dev Withdraw assets using PalletXCM call.
+     * @return A boolean confirming whether the XCM message sent.
+     */
```
Dinonard: It would be good to provide better documentation for the teams.

This will the main thing they'll be looking at most likely.

precompiles/xcm/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+        if fee_asset_item as usize > assets.len() {
+            return Err(gasometer.revert("bad fee index"));
+        }
+
+        // Prepare pallet-xcm call arguments
+        let dest = if is_relay {
+            MultiLocation::parent()
+        } else {
+            X1(Junction::Parachain(parachain_id)).into()
+        };
```
Dinonard, line 124: Tbh, I have some trouble wrapping my head around how useful is this part.

Based on the provided asset Ids, we know we exactly where we need to deposit them back.
If we don't, it means the asset mapping is missing and something's wrong.

Would it be valid to just take the first destination location derived from asset Id and ensure that it's the same parachain for all assets in the list?

precompiles/xcm/Cargo.toml
akru: This is not pallet-xcm, it's XCM primitives. And yes, pallet-xcm referenced as `path` above.

Dinonard: Omg, sorry, my bad, not sure what I saw :sweat_smile:

precompiles/xcm/src/lib.rs
akru: Not sure, this location is message destination but assets inside could have different reserve locations. For example, KSM and custom statemine token will have different reserves.

precompiles/xcm/XCM.sol
akru: Agree, let me fix it

Dinonard (approved):

precompiles/xcm/XCM.sol
```diff
@@ -0,0 +1,29 @@
+pragma solidity ^0.8.0;
+
+/**
+ * @title XCM interface.
+ */
+interface XCM {
```
Maar-io, line 6: can you name it IXCM?
Usually the interfaces start with 'I'

precompiles/xcm/src/lib.rs
Dinonard: You're right, I forgot about how I did it in demo.

Westmint accepted `{parents: 1, interior: Here}` as _native_ currency.

PR #58, opened 2022-05-31
Cross-chain Asset Config

Dinonard:
**Pull Request Summary**

Adds pallet to handle mapping between local asset and cross-chain asset.
Also handles payment rate for foreign assets.

The main idea is to provide simple pallet without intertwining it with `pallet-assets` or reinventing that pallet.

Note that at the moment it's not possible to add genesis config because `MultiLocation` isn't serializable.
We should change this in `polkadot` repo.

**TODO**
- [x] team review
- [x] tests
- [x] benchmarks
- [x] callback for EVM code injection handling

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


frame/xc-asset-config/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+    pub type AssetIdToLocation<T: Config> =
+        StorageMap<_, Blake2_128Concat, T::AssetId, VersionedMultiLocation>;
+
+    /// Mapping from an asset type to an asset id.
+    /// Can be used when receiving a multilocation XCM message to retrieve
+    /// the corresponding asset in which tokens should me minted.
+    #[pallet::storage]
+    #[pallet::getter(fn asset_location_to_id)]
+    pub type AssetLocationToId<T: Config> =
+        StorageMap<_, Blake2_128Concat, VersionedMultiLocation, T::AssetId>;
```
Dinonard: TODO: Change to more efficient twox hashing.

Dinonard:
@akru @Maar-io @codespool please review/approve.

This has been integrated into Shibuya (locally) and tested extensively.

frame/xc-asset-config/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+    use parity_scale_codec::HasCompact;
+    use sp_std::boxed::Box;
+    use xcm::{v1::MultiLocation, VersionedMultiLocation};
+
+    #[pallet::pallet]
+    #[pallet::without_storage_info]
+    pub struct Pallet<T>(PhantomData<T>);
+
+    /// Callback definition trait for cross-chain asset registration/deregistration notifications.
+    pub trait XcAssetChangedCallback<T: Config> {
```
akru: I believe we could easily drop Callback from name without loss in means.

frame/xc-asset-config/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+        /// Will be called by pallet when new asset Id has been registered
+        fn xc_asset_registered(asset_id: T::AssetId);
+
+        /// Will be called by pallet when asset Id has been unregistered
+        fn xc_asset_unregistered(asset_id: T::AssetId);
+    }
+
+    /// Defines conversion between asset Id and asset type
+    pub trait AssetLocationGetter<AssetId> {
```
akru: Why don’t keep same approach like XcAssetLocation and xc_ prefix for methods?

frame/xc-asset-config/src/lib.rs
Dinonard: Sure.

frame/xc-asset-config/src/lib.rs
Dinonard: Sure, I can change it.

akru (approved):
LGatM

PR #62, opened 2022-06-16
Feature/chain extension dapps staking

Maar-io:
**Pull Request Summary**

> Adds chain-extension for dapps-staking. Used by WASM contracts

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)


chainextensions/dapps-staking/src/lib.rs
```diff
@@ -0,0 +1,90 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+use sp_runtime::DispatchError;
+
+use codec::Encode;
+use frame_support::log::trace;
+use pallet_contracts::chain_extension::{Environment, Ext, InitState, SysConfig, UncheckedFrom};
+use pallet_dapps_staking;
+
+pub trait AstarChainExtension {
```
shunsukew: How about more general naming such as `ChainExtension` or `ChainExtensionExec`? This chain extension can be used in different runtimes (not only Astar) as long as it implements dapps-staking and contract pallets.

We can avoid naming conflicts (with pallet-contracts' ChainExtension trait) by using rename feature on runtime side.
`pub use pallet_chain_extension_dapps_staking::{ChainExtension as AstarChainExtension, DappsStakingExtension};`

chainextensions/dapps-staking/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn execute_func<E, R>(
+        func_id: u32,
+        env: Environment<E, InitState>,
+    ) -> Result<(), DispatchError>
+    where
+        E: Ext,
+        <E::T as SysConfig>::AccountId: UncheckedFrom<<E::T as SysConfig>::Hash> + AsRef<[u8]>,
+        R: pallet_dapps_staking::Config;
```
shunsukew: (1) In my opinion, pallets' Config trait better not to be used as trait bound here.
Because the more pallets support, the more tightly coupled with a particular runtime. Finally, we will have the situation where this trait is available for runtime X (pallet_a, pallet_b, pallet_c) but not available for runtime Y (pallet_a, pallet_b, pallet_d).
dapps staking chain extension should be available no matter what combination of pallets runtime has. (As long as it implements dapps-staking pallet)

(2) If you look at `Ext` trait, it contains `pallet_contracts::Config` that means `frame_system:Config` trait.
```
pub trait Ext: sealing::Sealed {
	type T: Config;
```
So, we can see Config trait duplication between `Ext::T` and `R`.

One of ideas to avoid avobe is here https://github.com/AstarNetwork/astar-frame/pull/65/files

Dinonard: @shunsukew 
I'm not sure I understand the scenario under 1).
This chain extension is for `pallet_dapps_staking` so having it depend on that pallet's config isn't problematic.
There is no mention of runtimes here, only a type `R` that is bound to `pallet_dapps_staking::Config`.

I'm also not sure what you mean by _the more pallet support, the more tightly coupled with a particular runtime_. Can you please explain?

shunsukew: If my understanding is correct, `AstarChainExtension` trait is for enforcing the same behavior `execute_func` to all chain extensions (like evm Precompile trait does in [astar-frame](https://github.com/AstarNetwork/astar-frame/blob/polkadot-v0.9.24/precompiles/sr25519/src/lib.rs) & [runtime](https://github.com/AstarNetwork/Astar/blob/master/runtime/shibuya/src/precompiles.rs)).

That means, this `AstarChainExtension` trait will be exported somewhere at the time of next chainextension support (perhaps balances-pallet) and used to all chainextension implementation not only in dapps-staking.

> I'm also not sure what you mean by the more pallet support, the more tightly coupled with a particular runtime. Can you please explain?

I mean the trait `AstarChainExtension` will be implemented to all chainextensions not only `DappsStakingChainExtension` struct. If this is not the case, what is the purpose of having trait `AstarChainExtension`?

Dinonard: Oh, of course!

The general trait shouldn't be bound to any pallet in particular, completely correct :+1:

chainextensions/dapps-staking/src/lib.rs
```diff
@@ -0,0 +1,567 @@
...
+                return Err(DispatchError::Other(
+                    "DappsStakingExtension: Unimplemented func_id",
+                ))
+            }
+        }
+    }
+}
+
+// pub struct DappsStakingExtension;
+pub struct DappsStakingExtension<R>(PhantomData<R>);
```
shunsukew: What is the purpose of having `PhantomData`?

chainextensions/types/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "chain-extension-types"
```
shunsukew: What is the purpose of having types package?

My opinion is all chainextensions (including types) should be independent of each other. So, it's better to put types in separate packages, rather than all chainextensions' types in a single package.
Because different chains have a different combination of pallets. Substrate based chains should be able to pick one chainextension even if it doesn't have another supported in astar-frame.

shunsukew: ~~Rust will compile crate as a unit. If you put all types (dapps-stakings' types, pallet-a's types, pallet-b's types) in one single package, unnecessary types will be compiled too, that's not good.
Assume that Chain-X import pallet-a chainextension, it will compile dapps-staking types too even if it doesn't need them.~~
Sorry, perhaps smart enough compiler can eliminate dead codes.

PierreOssun: _it's better to put types in separate packages, rather than all chainextensions' types in a single package._
Totally agree - I'll create a dapps-staking-chain-extension-types crate and only expose the relevant types

chainextensions/dapps-staking/impl/Cargo.toml
```diff
@@ -0,0 +1,46 @@
+[package]
+name = "pallet-chain-extension-dapps-staking"
+version = "4.0.0"
```
Dinonard: Why 4.0.0? :thinking:

chainextensions/dapps-staking/impl/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+            11 => Ok(DappsStakingFunc::WithdrawUnbonded),
+            12 => Ok(DappsStakingFunc::ClaimStaker),
+            13 => Ok(DappsStakingFunc::ClaimDapp),
+            14 => Ok(DappsStakingFunc::SetRewardDestination),
+            15 => Ok(DappsStakingFunc::NominationTransfer),
+            _ => Err(DispatchError::Other(
+                "DappsStakingExtension: Unimplemented func_id",
+            )),
+        }
+    }
```
Dinonard: Why not adopt the same approach as EVM does regarding the function identifier?

chainextensions/dapps-staking/impl/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+        let func_id = DappsStakingFunc::try_from(func_id)?;
+        let mut env = env.buf_in_buf_out();
+
+        match func_id {
+            // DappsStaking - read_current_era()
+            DappsStakingFunc::CurrentEra => {
+                let era_index = pallet_dapps_staking::CurrentEra::<T>::get();
+                env.write(&era_index.encode(), false, None).map_err(|_| {
+                    DispatchError::Other(
+                        "[ChainExtension] DappsStakingExtension failed to write result",
```
Dinonard: Would be good to include which function failed exactly. Currently all have the same error message which isn't helpful.

chainextensions/dapps-staking/impl/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+    where
+        E: Ext<T = T>,
+        <E::T as SysConfig>::AccountId:
+            UncheckedFrom<<E::T as SysConfig>::Hash> + AsRef<[u8]> + From<[u8; 32]>,
+    {
+        let func_id = DappsStakingFunc::try_from(func_id)?;
+        let mut env = env.buf_in_buf_out();
+
+        match func_id {
+            // DappsStaking - read_current_era()
```
Dinonard: Nit-picky comment but all these comments are redundant:
```Rust
// DappsStaking - read_current_era()
 DappsStakingFunc::CurrentEra => {
     let era_index = pallet_dapps_staking::CurrentEra::<T>::get();
```

chainextensions/dapps-staking/impl/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+                            "[ChainExtension] DappsStakingExtension failed to write result",
+                        )
+                    })?;
+            }
+
+            // DappsStaking - read_era_reward()
+            DappsStakingFunc::EraRewards => {
+                let arg: u32 = env.read_as()?;
+                let era_info = pallet_dapps_staking::GeneralEraInfo::<T>::get(arg)
+                    .ok_or(DispatchError::Other("general_era_info call failed"));
```
Dinonard: Should `?` be added here?

chainextensions/dapps-staking/impl/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+                        DispatchError::Other(
+                            "[ChainExtension] DappsStakingExtension failed to write result",
+                        )
+                    })?;
+            }
+
+            // DappsStaking - read_contract_stake()
+            DappsStakingFunc::ReadContractStake => {
+                let contract_bytes: [u8; 32] = env.read_as()?;
+                let contract = Self::decode_smart_contract(contract_bytes)?;
```
Dinonard: This is somehow limiting - what if WASM contract wants to access EVM contract?
Since we're aiming to be a X-VM chain, isn't it important to consider that?

chainextensions/dapps-staking/impl/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+                    DispatchError::Other(
+                        "[ChainExtension] DappsStakingExtension failed to write result",
+                    )
+                })?;
+            }
+
+            // DappsStaking - register()
+            DappsStakingFunc::Register => {
+                let contract_bytes: [u8; 32] = env.read_as()?;
+                let contract = Self::decode_smart_contract(contract_bytes)?;
```
Dinonard: If this isn't updated to support EVM (and other contracts), you should consider implementing `From<[u8; 32]>` for the smart contract type and just use `into()`.

chainextensions/dapps-staking/impl/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+
+                let base_weight = <T as pallet_dapps_staking::Config>::WeightInfo::register();
+                env.charge_weight(base_weight)?;
```
Dinonard: Shouldn't the previous calls where only read is performed also charge weight for each read operation?

chainextensions/dapps-staking/traits/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "chain-extension-traits"
+version = "4.0.0"
```
Dinonard: Two questions:
* one repeated but why 4.0.0?
* `chain-extension-traits` seem to be very generic but they're under _chainextension/dapps-staking_. Why?

chainextensions/dapps-staking/traits/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+use pallet_contracts::chain_extension::{Environment, Ext, InitState, SysConfig, UncheckedFrom};
+use sp_runtime::DispatchError;
+
+pub trait ChainExtensionExec<T: SysConfig> {
```
Dinonard: Doc would be good.

chainextensions/dapps-staking/types/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "chain-extension-types"
```
Dinonard: Same questions as for `traits` TOML file.

Dinonard (commented):
Great work!

A few general questions:
* why `chainextension` and not `chain-extension`? Seems more in line with every other folder we have.
* no unit tests?

chainextensions/dapps-staking/impl/src/lib.rs
PierreOssun: @Dinonard can you share the code line of it?

chainextensions/dapps-staking/impl/src/lib.rs
PierreOssun: yes it returns a Result
https://github.com/paritytech/substrate/blob/ef890f5598e0d3fd1b6a5fcc8c3322f7570089d1/frame/contracts/src/chain_extension.rs#L369

chainextensions/dapps-staking/impl/src/lib.rs
PierreOssun: No because if you don't modify the on-chain storage then it is a query (from your smart-contract)
And queries are free of gas

chainextensions/dapps-staking/traits/Cargo.toml
PierreOssun: Yes the Trait will be used by any impls of chain-extensions
I refactored the folder structure to meet future implementations

chainextensions/dapps-staking/impl/src/lib.rs
Dinonard: Something like [this](https://github.com/AstarNetwork/astar-frame/blob/polkadot-v0.9.27/precompiles/utils/macro/src/lib.rs#L88)

chainextensions/dapps-staking/impl/src/lib.rs
Dinonard: I'm not sure that's correct when you're reading on-chain storage from a pallet.

Each read means accessing the DB and they are not free.
E.g. we're currently using RocksDB and each DB read withing extrinsic execution costs 25_000_000 units of weight while each write costs 100_000_000 units of weight. Since smart contract execution is an extrinsic and it accesses off-chain DB, it should also count any read it does as 'gas consumed'.

Same is true for EVM.

chainextensions/dapps-staking/impl/src/lib.rs
PierreOssun: @Dinonard I'll do it in another PR as there is a lot of changes to do this.
I need to pass the SmartCotnract enum instead of a 32 bytes array

chainextensions/dapps-staking/impl/src/lib.rs
PierreOssun: I'll fix it in when Implementing staking for EVM contract as well

chainextensions/dapps-staking/impl/src/lib.rs
Dinonard: Sure, let's just make sure we don't deploy on Shiden before that is deployed & tested on Shibuya :+1:

chain-extensions/impls/dapps-staking/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+                        .unwrap_or_default();
+                let total = TryInto::<u128>::try_into(staking_info.total).unwrap_or(0);
+                env.write(&total.encode(), false, None).map_err(|_| {
+                    DispatchError::Other(
+                        "[ChainExtension] DappsStakingExtension ReadContractStake failed to write result",
+                    )
+                })?;
+            }
+
+            DappsStakingFunc::Register => {
```
Dinonard: We will deprecate this call soon (will be made root-only) so I'd suggest not adding it at all.

chain-extensions/impls/dapps-staking/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+                    }
+                };
+            }
+
+            DappsStakingFunc::ClaimStaker => {
+                let contract_bytes: [u8; 32] = env.read_as()?;
+                let contract = Self::decode_smart_contract(contract_bytes)?;
+                let base_weight = T::WeightInfo::claim_staker_with_restake()
+                    .max(T::WeightInfo::claim_staker_without_restake());
+                env.charge_weight(base_weight)?;
```
Dinonard: I believe most users are using automatic restake option and in that case the weight of the call is higher. So you should first use `T::WeightInfo::claim_staker_with_restake`.

After receiving the result from executing the extrinsic call (line 305), you should fetch from it the real used weight and call `adjust_weight` on `env`.

Dinonard (approved):
LGTM :+1:

PierreOssun (approved):

chain-extensions/impls/dapps-staking/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+    ) -> Result<<R as pallet_dapps_staking::Config>::SmartContract, DispatchError>
+    where
+        R: pallet_dapps_staking::Config,
+        R::AccountId: From<[u8; 32]>,
+    {
+        let account: R::AccountId = contract_bytes.into();
+        // Encode contract address to fit SmartContract enum.
+        // Since the SmartContract enum type can't be accessed from this chain extension,
+        // use locally defined enum clone (see Contract enum)
+        let contract_enum_encoded = Contract::<R::AccountId>::Wasm(account).encode();
```
shunsukew, line 462: Input address is not necessarily Wasm AccountId. It can be EVM H160.
Is this taken into consideration?
Or only AccountId is allowed for wasm contracts?

PR #65, opened 2022-06-27
chain extension balances

shunsukew:
This PR will be closed and migrated to official PR which will be created later.
Sharing design idea purpose.


PierreOssun:
@shunsukew @Dinonard I think we can close this

PR #68, opened 2022-07-01
XC20+

Dinonard:
**Pull Request Summary**

**WIP**

This PR extends `XC20` interface by additional functions.
We call this new interface `XC20+`.

New functions:
* minimum_balance
* mint
* burn

**Check list**
- [x] added unit tests
- [x] updated documentation
- [x] test locally

### DIFF
--- a/precompiles/assets-erc20/ERC20.sol
+++ b/precompiles/assets-erc20/ERC20.sol
@@ -108,4 +108,22 @@
         address indexed spender,
         uint256 value
     );
+
+    /**************************/
+    /********** XC20+ *********/
+    /**************************/
+
+    /**
+     * @dev Returns minimum balance an account must have to exist
+     * Selector: b9d1d49b
+     */
+    function minimumBalance() external view returns (uint256);
+
+    /**
+     * @dev Mints the specified amount of asset for the beneficiary.
+     * This operation will increase total supply.
+     * Only usable by asset owner (issuer).
+     * Selector: 40c10f19
+     */
+    function mint(address beneficiary, uint256 amount) external view returns (bool);
 }
--- a/precompiles/assets-erc20/ERC20.sol
+++ b/precompiles/assets-erc20/ERC20.sol
@@ -128,7 +128,7 @@
      * Only usable by asset admin.
      * Selector: 40c10f19
      */
-    function mint(address beneficiary, uint256 amount) external view returns (bool);
+    function mint(address beneficiary, uint256 amount) external returns (bool);
 
     /**
      * @dev Burns by up to the specified amount of asset from the target.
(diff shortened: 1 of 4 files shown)


precompiles/assets-erc20/ERC20.sol
```diff
@@ -1,11 +1,11 @@
  pragma solidity ^0.8.0;
 
     /**
-     * @title ERC20 interface
+     * @title ERC20 interface extended by additional methods
      * @dev see https://github.com/ethereum/EIPs/issues/20
      * @dev copied from https://github.com/OpenZeppelin/openzeppelin-contracts
      */
-    interface IERC20 {
+    interface IERC20Plus {
```
akru: Let’s keep IERC20 as is, it’s Ethereum standard, just add one more interface with new name.

Dinonard: Added new interface in the same file and made it extend the original.
That way smart contracts can just use the original or the extended one, they don't need both.

akru (approved):

precompiles/assets-erc20/ERC20.sol
```diff
@@ -109,3 +109,33 @@
...
+     */
+    function minimumBalance() external view returns (uint256);
+
+    /**
+     * @dev Mints the specified amount of asset for the beneficiary.
+     * This operation will increase the total supply.
+     * Only usable by asset admin.
+     * Selector: 40c10f19
+     */
+    function mint(address beneficiary, uint256 amount) external returns (bool);
```
Maar-io, line 131: in the OpenZeppelin's [reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol) implementation both `mint` and `burn` are `internal` function, therefore can only be called within the smart contract and derived contracts

Dinonard: OpenZeppelin isn't referenced for `IERC20Plus`, only for the original.

Maar-io: And since they are internal, they are named `_mint` and `_burn`

precompiles/assets-erc20/ERC20.sol
Maar-io: Maybe plus is not best name. Something like `xc20mintable` sounds more explanatory

precompiles/assets-erc20/ERC20.sol
Dinonard: Ok, but can you please explain what you're trying to say?

Because I don't see any impediments for merge.

Maar-io (approved):
LGTM

precompiles/assets-erc20/ERC20.sol
Dinonard: Imo, it's good since it doesn't tie it to a specific functionality.
There's also the `minimum_balance` which isn't related to `mintable`.

PR #69, opened 2022-07-03
Feature/dapps staking root only registration

shunsukew:
**Pull Request Summary**
Dapps staking root-only contract registration.

Right now, when devs register a dapp for dapps-staking, they first need to be whitelisted. But only their account is whitelisted, we don’t know exactly which dapp they will deploy.
If we make the entire process atomic without `developer_pre_approval` process, whoever wants to register can specify exactly what the dev account is and which contract will be used.
Also, this is required for governance over dapps registration, because RootOrigin call will be scheduled to `pallet-scheduler` by `pallet-democracy` once referendums passed.

**Check list**
- [x] added unit tests
- [ ] updated documentation

Logs of try-runtime against Shibuya
Screen Shot 2022-07-10 at 1 28 20


shunsukew:
next pallet version could be `V3_1_0`, not `V4_0_0`

frame/dapps-staking/src/lib.rs
```diff
@@ -83,10 +81,7 @@ pub type BalanceOf<T> =
 pub type EraIndex = u32;
 
 /// Simple trait used to check whether the underlying struct represents a valid smart contract on-chain.
```
Dinonard: Docs should also be updated, now it's just a wrapper.

Perhaps the trait bounds from the pallet config could just be put here?
Especially the part about encoding.

frame/dapps-staking/src/tests.rs
```diff
@@ -358,6 +358,21 @@ fn register_is_ok() {
     })
 }
 
+#[test]
+fn register_with_non_root_fails() {
```
Dinonard, line 362: Great!

precompiles/dapps-staking/src/lib.rs
```diff
@@ -175,22 +175,9 @@ where
...
-        log::trace!(target: "ds-precompile", "register {:?}", contract_id);
-
-        // Build call with origin.
-        let origin = R::AddressMapping::into_account_id(handle.context().caller);
-        let call = pallet_dapps_staking::Call::<R>::register { contract_id }.into();
-
-        RuntimeHelper::<R>::try_dispatch(handle, Some(origin).into(), call)?;
-
-        Ok(succeed(EvmDataWriter::new().write(true).build()))
+    fn register(_: &mut impl PrecompileHandle) -> EvmResult<PrecompileOutput> {
```
Dinonard, line 179: Please update the `.sol` file too.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -332,6 +333,85 @@ pub mod pallet {
...
+                log::info!(">>> PreApprovalDevelopers cleanup stopped due to reaching max amount of deletions.");
+                consumed_weight += deletion_weight * removed_entries_num as u64;
+                if cfg!(feature = "try-runtime") {
+                    return consumed_weight + cleanup_pre_approved_developers::<T>();
+                } else {
+                    return consumed_weight;
+                }
+            }
+        }
     }
```
Dinonard: This isn't needed since this cleanup is lightweight.

Just add `PreApprovedDevelopers::<T>::remove_all(None)` to the original call and it will be fine.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -332,6 +333,85 @@ pub mod pallet {
...
+            );
+
+            assert_eq!(Version::V3_0_0, StorageVersion::<T>::get());
+
+            Ok(())
+        }
+
+        fn on_runtime_upgrade() -> Weight {
+            // Maintenance mode to prevent existing extrinsics update PreApprovedDevelopers map.
+            PalletDisabled::<T>::put(true);
```
Dinonard: You don't need to do this since you'll do cleanup in one step.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -332,6 +333,85 @@ pub mod pallet {
...
+                current_pre_approved_developers_count
+            );
+
+            assert_eq!(Version::V3_0_0, StorageVersion::<T>::get());
+
+            Ok(())
+        }
+
+        fn on_runtime_upgrade() -> Weight {
+            // Maintenance mode to prevent existing extrinsics update PreApprovedDevelopers map.
```
Dinonard: This isn't possible since you'll execute entire cleanup atomically.

Dinonard (commented):
Great work Shunsuke!
Very impressive for a side-task :smile:

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -190,11 +191,13 @@ pub mod pallet {
         false
     }
 
+    /// Deprecated. Need to be cleaned up with OnRuntimeUpgrade
     /// Enable or disable pre-approval list for new contract registration
     #[pallet::storage]
     #[pallet::getter(fn pre_approval_is_enabled)]
     pub(crate) type PreApprovalIsEnabled<T> = StorageValue<_, bool, ValueQuery, PreApprovalOnEmpty>;
```
Dinonard, line 199: Make sure to also `kill` this storage value.

precompiles/dapps-staking/src/lib.rs
shunsukew: The interface itself has no changes. Do you mean adding comment to .sol file too? register is root-origin call, it should always fail when called via evm precompile or something like that

frame/dapps-staking/src/pallet/mod.rs
shunsukew: Practically, this is good because they are lightweight?
```
fn on_runtime_upgrade() -> Weight {
    // lightweight deletions
    StorageVersion::<T>::put(Version::default());
    PreApprovalIsEnabled::<T>::kill();
    PreApprovedDevelopers::<T>::remove_all(None);
    0
 }
```

Dinonard: Yes, it's relatively lightweight but that code snippet is incorrect.

You should still return the consumed weight.

1 write + 1 write + `N` writes.
`remove_all` will return an enum containing the amount of removed entries so you can use that.

precompiles/dapps-staking/src/lib.rs
Dinonard: Yes, I mean the comments.
It should be clear that this is a deprecated functionality and shouldn't be used anymore.

frame/dapps-staking/src/pallet/mod.rs
shunsukew: I understand what you meant. thank you!

Dinonard:
One thing to also consider - the `weights.rs` are no longer valid since you've changed the interface.
That should be updated.

We should re-run the benchmarks but we don't have a dedicated benchmarking machine.
I'd suggest you just remove the weight consumed by pre-approval checks that have now been removed.

shunsukew:
> One thing to also consider - the `weights.rs` are no longer valid since you've changed the interface. That should be updated.
> 
> We should re-run the benchmarks but we don't have a dedicated benchmarking machine. I'd suggest you just remove the weight consumed by pre-approval checks that have now been removed.

I know my local machine is weaker than dedicated benchmarking machine satisfying requirements. but local execution had the same result this time.
```
========
Storage: DappsStaking PalletDisabled (r:1 w:0)
Storage: DappsStaking RegisteredDevelopers (r:1 w:1)
Storage: DappsStaking RegisteredDapps (r:1 w:1)
```

How did you define base weight `32_139_000 `?

frame/dapps-staking/src/lib.rs
shunsukew: I tried to inherit `Parameter` and `Member` for `IsContract` trait, but this is not possible. `frame_support::traits::dispatch::{Parameter, Member}`
`dispatch` is private module and not accessible directly. I assume `Parameter` and `Member` are available in `#[pallet::config]`

I think no need to have `IsContract` trait anymore, although it requires changes on Runtime side as well if we remove it. and adding Default to trait boundary. Is it okay?

Dinonard:
> > One thing to also consider - the `weights.rs` are no longer valid since you've changed the interface. That should be updated.
> > We should re-run the benchmarks but we don't have a dedicated benchmarking machine. I'd suggest you just remove the weight consumed by pre-approval checks that have now been removed.
> 
> I know my local machine is weaker than dedicated benchmarking machine satisfying requirements. but local execution had the same result this time.
> 
> ```
> ========
> Storage: DappsStaking PalletDisabled (r:1 w:0)
> Storage: DappsStaking RegisteredDevelopers (r:1 w:1)
> Storage: DappsStaking RegisteredDapps (r:1 w:1)
> ```
> 
> How did you define base weight `32_139_000 `?

I didn't define it, it was measured by benchmarking tool.

The part you pasted would be same on any machine, it's discrete measurement of number of DB operations.

It's strange that only one read is removed since the two reads for PreApproval have been removed.
But I see now that benchmarking was updated to take into consideration both branches - one with pre-approval enabled, one without.
Anyhow, the modification you did is fine. :+1:

shunsukew:
> > > One thing to also consider - the `weights.rs` are no longer valid since you've changed the interface. That should be updated.
> > > We should re-run the benchmarks but we don't have a dedicated benchmarking machine. I'd suggest you just remove the weight consumed by pre-approval checks that have now been removed.
> > 
> > 
> > I know my local machine is weaker than dedicated benchmarking machine satisfying requirements. but local execution had the same result this time.
> > ```
> > ========
> > Storage: DappsStaking PalletDisabled (r:1 w:0)
> > Storage: DappsStaking RegisteredDevelopers (r:1 w:1)
> > Storage: DappsStaking RegisteredDapps (r:1 w:1)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > How did you define base weight `32_139_000 `?
> 
> I didn't define it, it was measured by benchmarking tool.
> 
> The part you pasted would be same on any machine, it's discrete measurement of number of DB operations.
> 
> It's strange that only one read is removed since the two reads for PreApproval have been removed. But I see now that benchmarking was updated to take into consideration both branches - one with pre-approval enabled, one without. Anyhow, the modification you did is fine. 👍

I see.
Oh, I wonder why I thought so... Yes this will calculate actual weight it costs.
```
T::DbWeight::get().reads(3 as Weight)
```

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -333,6 +336,58 @@ pub mod pallet {
...
+            StorageVersion::<T>::put(Version::default());
+            PreApprovalIsEnabled::<T>::kill();
+
+            let deletion_weight = T::DbWeight::get().writes(1) * 11 / 10;
+            match PreApprovedDevelopers::<T>::remove_all(None) {
+                KillStorageResult::AllRemoved(removed_entries_num) => {
+                    consumed_weight += deletion_weight * removed_entries_num as u64;
+                }
+                KillStorageResult::SomeRemaining(removed_entries_num) => {
+                    consumed_weight += deletion_weight * removed_entries_num as u64;
```
Dinonard: I'd suggest doing an error trace instead - this mustn't happen.

Dinonard (approved):
LGTM, well done!

Please coordinate with UI team before the merge.
There's plan to upgrade deps to `polkadot-v0.9.26` so you might want to wait until that before merging your PR.

frame/dapps-staking/src/pallet/mod.rs
shunsukew: @Dinonard It seems, with polkadot-0.9.26, some changes are introduced.
1. `remove_all` has been deprecated, `clear` instead.
2. `clear` will return simple struct which includes the number of unique keys removed.

So, I updated corresponding part `on_runtime_upgrade` function. Please check.

shunsukew:
> next pallet version could be `V3_1_0`, not `V4_0_0`

We broke `register` API backward compatibility, so `v4_0_0` should be fine.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Looks good :+1:

PR #72, opened 2022-07-11
ChainExtension: Unit test for Dapps-staking

Maar-io:
**Check list**
- [ ] added unit tests
- [ ] updated documentation


Maar-io:
Canceled
Did not find the way to use unit test

PR #73, opened 2022-07-12
Ci improvements

shunsukew:
**Pull Request Summary**
CI improvements.
This requires code changes to solve lint and format checks.

- added `taplo` Cargo.toml formatter.
- added clippy lint check
  some checks are configured to be ignored at `.cargo/config.toml`


.github/workflows/checks-and-tests.yml
```diff
@@ -23,7 +16,7 @@ jobs:
     - name: Check if code compiles
       run: cargo check --release --all-features --verbose
 
-  test-and-coverage:
+  tests:
```
Dinonard, line 34: Can we use cache here?

Also, there's the `tarpaulin` below which we use to generate coverage report.
But maybe there's a better approach to this?

I believe I saw some project using UT coverage and the report was published as a comment inside review section.

.github/workflows/checks-and-tests.yml
```diff
@@ -1,18 +1,11 @@
...
-
+name: Checks and Tests
+on: [push, workflow_dispatch]
 jobs:
-  native:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
+  checks:
```
Dinonard, line 7: Can we use cache here?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -22,6 +22,7 @@ use sp_std::convert::From;
 const STAKING_ID: LockIdentifier = *b"dapstake";
 
 #[frame_support::pallet]
+#[allow(clippy::module_inception)]
```
Dinonard, line 25: Imo, we should fix this.
It's been a thorn in my eye for quite some time.

But to keep this task focused, can you crate a new backlog item for this?
It should be low prio, P5.

Dinonard (commented):
Nice, good work!

Just a few small comments, otherwise should be ready for merge.

.github/workflows/checks-and-tests.yml
```diff
@@ -0,0 +1,62 @@
+name: Checks and Tests
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+  workflow_dispatch:
+jobs:
+  checks-and-tests:
```
Dinonard: Sorry but I got more questions :see_no_evil: 

I see that you've bundled together 3 actions into one now.
Won't this make the overall process take longer?
We could run UTs and build in parallel, right?

It only takes a few minutes anyhow.

shunsukew: It seems parallel is faster
In pararell, `6m 59s` for test and `4m 8s` for checks. -> `6m 59s`
In sequential, `5m 13s` for test and `3m 47s` for checks. -> `9m`

let's roll it back!

frame/dapps-staking/src/pallet/mod.rs
shunsukew: I wonder why we have pallet directory at first sight. but It's easy to fix this, by just changing directory name or pub mod name. What it could be. I suggest folder name `impl`?

Or we can flatten and put everything inside `lib.rs`

Dinonard: We don't need that directory at all.

But don't change it as part of this task, this is a breaking change for runtimes.

.github/workflows/checks-and-tests.yml
```diff
@@ -1,5 +1,8 @@
 name: Checks and Tests
-on: [push, workflow_dispatch]
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+  workflow_dispatch:
```
Dinonard: This will skip drafts?

Dinonard (approved):
LGTM

.github/workflows/checks-and-tests.yml
shunsukew: actually no. I added if statement

Dinonard: Oh I see, I guess that's fine.

Originally I was just interested in what the types meant :smile:

PR #74, opened 2022-07-12
XVM pallet MVP v1

akru:
**Pull Request Summary**

This PR implements v1 XVM pallet.

**Check list**
- [ ] added unit tests
- [ ] updated documentation

### DIFF
--- a/frame/pallet-xvm/src/lib.rs
+++ b/frame/pallet-xvm/src/lib.rs
@@ -1,0 +1,174 @@
+//! # XVM pallet
+//!
+//! ## Overview
+//!
+//!
+//! ## Interface
+//!
+//! ### Dispatchable Function
+//!
+//!
+//! ### Other
+//!
+//!
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use codec::{Decode, Encode};
+use sp_runtime::{traits::Member, RuntimeDebug};
+use sp_std::prelude::*;
+
+pub mod pallet;
+pub use pallet::pallet::*;
+
+/// EVM call adapter instance.
+#[cfg(feature = "evm")]
+pub mod evm;
+
+/// Wasm call adapter instance.
+#[cfg(feature = "wasm")]
+pub mod wasm;
+
+/// XVM context consist of unique ID and optional execution arguments.
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, scale_info::TypeInfo)]
+pub struct XvmContext<VmId> {
+    /// VM identifier.
+    pub id: VmId,
+    /// Encoded VM execution environment.
+    pub env: Option<Vec<u8>>,
+}
+
+/// The engine that support synchronous smart contract execution.
+/// For example, EVM.
+///
+/// An impl code should realize input data conversion using provided
+/// metadata.
+pub trait SyncVM<VmId, AccountId> {
+    /// Unique VM identifier.
+    fn id() -> VmId;
+
+    /// Make a call to VM contract and return result or error.
+    ///
+    ///
+    fn xvm_call(
+        context: XvmContext<VmId>,
+        from: AccountId,
+        to: Vec<u8>,
+        input: Vec<u8>,
+        metadata: Vec<u8>,
+    ) -> Result<Vec<u8>, Vec<u8>>;
+}
+
+#[impl_trait_for_tuples::impl_for_tuples(30)]
+impl<VmId: Member + Default, AccountId: Member> SyncVM<VmId, AccountId> for Tuple {
+    fn id() -> VmId {
+        Default::default()
+    }
+
+    fn xvm_call(
+        context: XvmContext<VmId>,
+        from: AccountId,
+        to: Vec<u8>,
+        input: Vec<u8>,
+        metadata: Vec<u8>,
+    ) -> Result<Vec<u8>, Vec<u8>> {
+        for_tuples!( #(
+            if Tuple::id() == context.id {
+                log::trace!(
+                    target: "xvm::SyncVm::xvm_call",
+                    "VM found, run XVM call: {:?}, {:?}, {:?}, {:?}, {:?}",
+                    context, from, to, input, metadata,
+                );
+                return Tuple::xvm_call(context, from, to, input, metadata)
+            }
+        )* );
+        log::trace!(
+            target: "xvm::SyncVm::xvm_call",
+            "VM with ID {:?} not found", context.id
+        );
+        Err(b"VM is not found".to_vec())
+    }
+}
+
+/// The engine that support asynchronous smart contract execution.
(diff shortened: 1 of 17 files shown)


frame/pallet-xvm/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+        Default::default()
+    }
+}
+
+/// Universal VM parameter codec.
+///
+/// Input data is SCALE encoded, output is specified for each VM.
+pub trait XvmCodec {
+    /// Convert generic XVM input into VM native.
+    fn encode(input: Vec<u8>, metadata: Vec<u8>) -> Vec<u8>;
```
Dinonard: I believe this should be `decode`?

`encode` should return encoded data and metadata, right?

precompiles/xvm/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+        From<pallet_xvm::Call<R>> + Dispatchable<PostInfo = PostDispatchInfo> + GetDispatchInfo,
+{
+    fn xvm_call(
+        handle: &mut impl PrecompileHandle,
+    ) -> EvmResult<PrecompileOutput> {
+        let mut input = handle.read_input()?;
+        input.expect_arguments(4)?;
+
+        // Read arguments and check it
+        let context_raw = input.read::<Bytes>()?;
```
Dinonard, line 64: I must admit that I don't understand what the context is here?

What extras should the smart contract provide? Why aren't parameters enough?

frame/pallet-xvm/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+#[cfg(feature = "wasm")]
+pub mod wasm;
+
+/// XVM context consist of unique ID and optional execution arguments.
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, scale_info::TypeInfo)]
+pub struct XvmContext<VmId> {
+    /// VM identifier.
+    pub id: VmId,
+    /// Encoded VM execution environment.
+    pub env: Option<Vec<u8>>,
```
Dinonard, line 38: We've had discussion about this in slack but I still don't see why this is Vec<u8>.

It is true that different VMs have different view of e.g. gas (EVM vs wasm), but we need to handle concrete limitations like `gas limit`.

E.g. for EVM, we have gas limit in EVM context.
So the precompiles should take note of what is the gas limit and how much gas so far has been consumed.
We'd have a struct like:

```Rust
pub EvmExecutionInfo {
  gas_limit: U256,
  used_gas: U256,
  // and more data that is needed
}
```

There would also be a root XVM execution info representation like:
```Rust
pub struct XvmExecutionInfo {
    gas_limit: U256,
    used_gas: U256,
    // and more data that is needed
}
```

And we'd need to convert from one type to another:
```Rust
let evm_exec_info = ....;

let xvm_exec_info: XvmExecutionInfo = evm_exec_info.into();

let parity_wasm_exec_info: ParityWasmExecutionInfo = xvm_exec_info.into();
```
Such conversions should ensure that we keep track of the overall gas limit and how much was used during the calls. As I noted [here](https://github.com/AstarNetwork/astar-frame/pull/81/files#diff-31e0ebbf868cdd39dfea4c054957b66ff5f238240906269a1695a714c8c9b776R95), we need to pass this information around and keep it consistent.

Please correct me if my understanding is incorrect.

frame/pallet-xvm/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// Unique VM identifier.
+    fn id() -> VmId;
+
+    /// Make a call to VM contract and return result or error.
+    ///
+    ///
+    fn xvm_call(
+        context: XvmContext<VmId>,
+        from: AccountId,
+        to: Vec<u8>,
```
Dinonard, line 56: What is `to`? Does it combine address and function identifier?

frame/pallet-xvm/src/lib.rs
akru: Encode to native, decode from generic, so, it’s same meaning. May be better use `convert` here.

precompiles/xvm/src/lib.rs
akru: It’s XVM execution context: vmid + environment (gas, value, etc.)

frame/pallet-xvm/src/lib.rs
akru: VM adapter should care about it, not precompile, not xvm don’t known how VM deal with context and parameters.

akru: If context wrong - pallet will return error to client, no need to double check it - first on precompile and then again on adapter.

frame/pallet-xvm/src/lib.rs
akru: Just destination ID, may be address, may be MultiAddress, may be MultiLocation

precompiles/xvm/src/lib.rs
Dinonard: Ok, then my understanding was correct only I assumed that precompiles/chain-extensions would handle that.
We have a similar discussion (or same) in my PR.

frame/pallet-xvm/src/lib.rs
Dinonard: Not sure how what I proposed above disagrees with what you wrote :thinking: 

My suggestion wasn't that e.g. EVM preocompile ensures anything except two things:
* it constructs `EvmExecInfo` and passes it to `pallet-xvm`
* it updates the used gas after execution finishes

I guess you want the second point to be handled by VM adapter?

frame/pallet-xvm/src/lib.rs
Dinonard: Do you mean that depending on the `VmId`, smart contract will have to encode destination differently?
No common type which XVM converts to appropriate type for the destination VM?

frame/pallet-xvm/src/lib.rs
akru: Honestly, I’m not sure that EVM precompile can correctly parse/construct all possible XVM contexts without double functionality with XVM adapters.
But yes, we should somehow return consumed gas information to EVM, anyway precompile could get weight from XVM call dispatch info.

frame/pallet-xvm/src/lib.rs
akru: Yes, depend of VM destination could be different. May be something like multi-location is more concrete but generic way of destination encoding. Not sure I want bring XCM only for this primitive.

akru: It’s open question anyway, not decided yet how better to present it.

Dinonard (approved):
LGTM!

Let's get V1 merged and build-up V2!

PR #75, opened 2022-07-14
Recover Erc20AssetsPrecompileSet new function

shunsukew:
**Pull Request Summary**
new function was failing lint check, thus it was removed in the former PR.
Instead, Default derive was added.
https://github.com/AstarNetwork/astar-frame/pull/73/files#diff-a57fe059e97a2348d92ca778296fda9d73e137ca1dd8692e164451509b62401dL180

However, Runtime side was using new function and cannot compile if dependency updated, so we should recover it. 

Screen Shot 2022-07-14 at 20 58 29

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -1,7 +1,7 @@
 name: Static Analysis
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


Dinonard (approved):
LGTM

PR #76, opened 2022-07-15
fixing wrong comment for claim_staker()

Maar-io:
Misleading comment for claim_staker in the Solidity interface file

**Check list**
- [ ] added unit tests
- [ ] updated documentation

### DIFF
--- a/precompiles/dapps-staking/DappsStaking.sol
+++ b/precompiles/dapps-staking/DappsStaking.sol
@@ -4,39 +4,42 @@ pragma solidity >=0.7.0;
 
 /// Interface to the precompiled contract on Shibuya/Shiden/Astar
 /// Predeployed at the address 0x0000000000000000000000000000000000005001
+/// For better understanding check the source code:
+/// repo: https://github.com/AstarNetwork/astar-frame
+/// code: precompiles/dapps-staking/src
 interface DappsStaking {
 
     // Storage getters
 
     /// @notice Read current era.
-    /// @return era, The current era
+    /// @return era: The current era
     function read_current_era() external view returns (uint256);
 
     /// @notice Read unbonding period constant.
-    /// @return period, The unbonding period in eras
+    /// @return period: The unbonding period in eras
     function read_unbonding_period() external view returns (uint256);
 
     /// @notice Read Total network reward for the given era
-    /// @return reward, Total network reward for the given era
+    /// @return reward: Total network reward for the given era
     function read_era_reward(uint32 era) external view returns (uint128);
 
     /// @notice Read Total staked amount for the given era
-    /// @return staked, Total staked amount for the given era
+    /// @return staked: Total staked amount for the given era
     function read_era_staked(uint32 era) external view returns (uint128);
 
     /// @notice Read Staked amount for the staker
-    /// @param staker in form of 20 or 32 hex bytes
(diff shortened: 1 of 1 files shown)


precompiles/dapps-staking/DappsStaking.sol
```diff
@@ -54,12 +54,16 @@ interface DappsStaking {
...
+    ///         In order to claim multiple eras, this call has to be called multiple times.
     ///         Staker account is derived from the caller address.
-    function claim_staker(address) external;
+    /// @param smart_contract smart contract address used for staking
+    function claim_staker(address smart_contract) external;
 
     /// @notice Claim one era of unclaimed dapp rewards for the specified contract and era.
-    function claim_dapp(address, uint128) external;
+    /// @param smart_contract smart contract address
+    /// @param era, The era to be claimed
```
Dinonard: `era, The era`?

Dinonard:
This is good - we should consider making an example contract with usage and add it to docs and our repo.
Since this is for devs, they will probably find it easier to understand when they see concrete usage.

precompiles/dapps-staking/DappsStaking.sol
```diff
@@ -54,12 +54,16 @@ interface DappsStaking {
     /// @notice Withdraw all funds that have completed the unbonding process.
     function withdraw_unbonded() external;
 
-    /// @notice Claim one era of unclaimed staker rewards for the specifeid contract.
+    /// @notice Claim earned staker rewards for the oldest unclaimed era.
```
Dinonard, line 60: Why not add a disclaimer on top that devs should read pallet rustdocs to better understand these functionalities?

Dinonard (changes requested):

precompiles/dapps-staking/DappsStaking.sol
Maar-io: it looks stupid here, but with syntax highlighter is ok. Let me rephrase it
Screenshot 2022-07-15 at 13 56 25

precompiles/dapps-staking/DappsStaking.sol
Maar-io: added  on top of the file

precompiles/dapps-staking/DappsStaking.sol
Maar-io: @Dinonard check now

precompiles/dapps-staking/DappsStaking.sol
```diff
@@ -4,39 +4,42 @@ pragma solidity >=0.7.0;
 
 /// Interface to the precompiled contract on Shibuya/Shiden/Astar
 /// Predeployed at the address 0x0000000000000000000000000000000000005001
+/// For better understanding check the source code:
+/// repo: https://github.com/AstarNetwork/astar-frame
+/// code: precompiles/dapps-staking/src
```
Dinonard: Actually, the comment was to reference the pallet code - so please change that before the merge.
We have detailed rustdoc there.

Dinonard (approved):

precompiles/dapps-staking/DappsStaking.sol
Maar-io: changed

PR #80, opened 2022-07-26
Support for AccountKey20 in XCM precompiles

Dinonard:
**Pull Request Summary**

Adds support for ETH-style address to the XCM-precompile's withdraw function.
Legacy interface remains unchanged, instead a new function is added, as recommended.

One drawback to the current solution is that both H160 and SS58 are encoded as 256 bits.
This can be problematic since we cannot detect if user incorrectly withdraw for SS58 with H160 as beneficiary.

**Check list**
- [x] added unit tests
- [x] updated documentation

### DIFF
--- a/precompiles/xcm/src/lib.rs
+++ b/precompiles/xcm/src/lib.rs
@@ -25,8 +25,8 @@ mod tests;
 #[precompile_utils::generate_function_selector]
 #[derive(Debug, PartialEq)]
 pub enum Action {
-    AssetsWithdrawSS58 = "assets_withdraw(address[],uint256[],bytes32,bool,uint256,uint256)",
-    AssetsWithdrawH160 = "assets_withdraw(address[],uint256[],address,bool,uint256,uint256)",
+    AssetsWithdrawNative = "assets_withdraw(address[],uint256[],bytes32,bool,uint256,uint256)",
+    AssetsWithdrawEvm = "assets_withdraw(address[],uint256[],address,bool,uint256,uint256)",
 }
 
 /// A precompile that expose XCM related functions.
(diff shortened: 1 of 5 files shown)


precompiles/xcm/src/lib.rs
```diff
@@ -51,11 +52,20 @@ where
...
+            Action::AssetsWithdrawSS58 => Self::assets_withdraw(handle, BeneficiaryType::SS58),
+            Action::AssetsWithdrawH160 => Self::assets_withdraw(handle, BeneficiaryType::H160),
         }
     }
 }
 
+/// The supported beneficiary account types
+enum BeneficiaryType {
+    /// 256 bit public key is expected
+    SS58,
```
akru: Why not AccountId32/AccountId20? SS58 is encoding style, not account type.

Dinonard: Sure, you're right.

I added it when I started and forgot to improve.

Dinonard: Done, please check.

precompiles/xcm/src/lib.rs
```diff
@@ -51,11 +52,24 @@ where
...
         }
     }
 }
 
+/// The supported beneficiary account types
+enum BeneficiaryType {
+    /// 256 bit (32 byte) public key
+    AccountId32,
+    /// 160 bit (20 byte) address is expected
+    AccountKey20,
```
akru: Technically it’s not public _key_, it’s id.

Dinonard: I know, hence the comment above.

However, I used `AccountKey20` to keep it consistent with `Junction` enum.

akru: So, what’s about to drop Id/Key from name, like “Account32 | Account20”?

Dinonard: Sure, let's do that.

akru (approved):

PR #81, opened 2022-08-01
Feature/chain extension for xvm pallet

Dinonard:
**Pull Request Summary**

Chain extension for XVM V1 sync VM call.

This is WIP since underlying pallet under heavy development is subject to change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,9 @@
 [workspace]
 members = [
 	"chain-extensions/impls/dapps-staking",
+	"chain-extensions/impls/xvm",
 	"chain-extensions/trait",
+	"chain-extensions/types/*",
 	"frame/block-reward",
 	"frame/collator-selection",
 	"frame/custom-signatures",
@@ -18,4 +20,6 @@ members = [
 	"precompiles/utils",
 	"precompiles/xcm",
 	"precompiles/xvm",
+
+	"contracts/*",
 ]
--- a/chain-extensions/impls/xvm/Cargo.toml
+++ b/chain-extensions/impls/xvm/Cargo.toml
@@ -1,0 +1,46 @@
+[package]
+name = "xvm-chain-extension"
+version = "0.1.0"
+authors = ["Stake Technologies <devops@stake.co.jp>"]
+edition = "2021"
+license = "Apache-2.0"
+homepage = "https://astar.network"
+repository = "https://github.com/AstarNetwork/astar-frame"
+description = "Chain extension for XVM"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false }
+frame-support = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.28", default-features = false }
+frame-system = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.28", default-features = false }
+log = "0.4.16"
+num-traits = { version = "0.2", default-features = false }
+pallet-contracts = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.28", default-features = false, features = ["unstable-interface"] }
+pallet-contracts-primitives = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.28", default-features = false }
+pallet-contracts-rpc-runtime-api = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.28", default-features = false }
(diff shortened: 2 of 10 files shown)


chain-extensions/xvm/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+            )),
+        }
+    }
+}
+
+/// Since we cannot read multiple variable length params, we need to resort to this approach. TODO: can it be improved?
+#[derive(Encode, Decode)]
+struct CustomParams {
+    /// virtual machine identifier
+    vm_id: u8,
```
akru: I suggest to use generic here and get precise type from runtime.

akru: It could be Decodable.

chain-extensions/xvm/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+                    metadata,
+                } = env.read_as_unbounded(env.in_len())?;
+                let vm_id: T::VmId = vm_id.into();
+
+                // TODO: rethink this? How do we get valid env in chain extension? Need to know which data to encode.
+                // gas limit, taking into account used gas so far?
+                let _origin_address = env.ext().address().clone();
+                let _value = env.ext().value_transferred();
+                let _gas_limit = env.ext().gas_meter().gas_left();
+                let xvm_context = XvmContext {
```
akru: Why don’t get context from smart contract?

Dinonard: Not sure I know what you mean exactly, sorry.

Can you explain a bit more?
How can I get context from smart contract?

I fetched some data that I believe might be relevant for the context in the few lines above.
However, we haven't decided how to encode that data yet, right?

chain-extensions/xvm/src/lib.rs
Dinonard: It already is decodable but I don't think this is the right approach given that we should make all types concrete to be used by the smart contract (cannot have `T::VmId`).

So we'll have to move such types to some crate with common smart contract types that are also shared by the chain extensions, to avoid ugly code duplication.

primitives/xvm/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use codec::{Decode, Encode};
+use scale_info::TypeInfo;
+
+// TODO: add RuntimeDebug? Maybe only use it when not building for runtime?
+/// Virtual machine identifier in the XVM context
+#[derive(PartialEq, Eq, Clone, Encode, Decode, TypeInfo)]
+pub enum VmId {
+    Unsupported,
```
akru: What’s about ‘Unsupported = n’ notation?

chain-extensions/xvm/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+                } = env.read_as_unbounded(env.in_len())?;
+
+                // TODO: rethink this? How do we get valid env in chain extension? Need to know which data to encode.
+                // gas limit, taking into account used gas so far?
+                let _origin_address = env.ext().address().clone();
+                let _value = env.ext().value_transferred();
+                let _gas_limit = env.ext().gas_meter().gas_left();
+                let xvm_context = XvmContext {
+                    id: vm_id,
+                    env: None,
```
akru: So, we have no way to pass custom environment from smart contract?

primitives/xvm/src/lib.rs
Dinonard: I'm not sure what you mean - haven't seen that notation with enums.

I only added this type because you require `Default` to be implemented over it when implementing tuples.
Personally, I'd remove it.

chain-extensions/xvm/src/lib.rs
Dinonard: We can pass anything as an argument, we'd just need to add it to the `CustomParams` struct above (and inside chain-extension so it can be decoded).

My understanding was that this `XvmContext` would be filled out by the chain extension, by fetching data from `Ext` as illustrated few lines above. It's VM related and I haven't seen smart contracts checking how much gas they've consumed so far.

Can you give me an example of custom environment that a smart contract would need to pass?

akru (approved):

chain-extensions/xvm/src/lib.rs
akru: Let's deal with it in next release.

PR #84, opened 2022-08-17
dApps-Staking chain-extension - Reworked Errors

PierreOssun:
Before the chain extension returned Result<() DispatchError> but it was not possible to decode it on contract side (in a handy way).
The appropriate way is to write to buffer when a value needs to be returned.
When a Result should be returned it should be wrapped in a RetVal and decoded on the contract side. This way no scale decoding error can happen.

**Check list**
- [X] Updated Trait function Return type
- [X] Updated all return statements

### DIFF
--- a/chain-extensions/types/dapps-staking/src/lib.rs
+++ b/chain-extensions/types/dapps-staking/src/lib.rs
@@ -65,8 +65,6 @@ pub enum DSError {
     NominationTransferToSameContract = 26,
     /// Unexpected reward destination value
     RewardDestinationValueOutOfBounds = 27,
-    /// Failed to write result on buffer
-    FailedToWriteOnBuffer = 28,
     /// Unknown error
     UnknownError = 99,
 }
--- a/chain-extensions/types/dapps-staking/src/lib.rs
+++ b/chain-extensions/types/dapps-staking/src/lib.rs
@@ -7,62 +7,68 @@ use sp_runtime::{DispatchError, ModuleError};
 #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
 #[derive(PartialEq, Eq, Copy, Clone, Encode, Decode, Debug)]
 pub enum DSError {
+    /// Success
+    Success = 0,
     /// Disabled
-    Disabled,
+    Disabled = 1,
     /// No change in maintenance mode
-    NoMaintenanceModeChange,
+    NoMaintenanceModeChange = 2,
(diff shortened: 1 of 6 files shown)


shunsukew (approved):

chain-extensions/types/dapps-staking/src/lib.rs
```diff
@@ -102,7 +106,7 @@ impl TryFrom<DispatchError> for DSError {
             Some("NominationTransferToSameContract") => {
                 Ok(DSError::NominationTransferToSameContract)
             }
-            _ => Err(DispatchError::Other("DappsStakingExtension: Unknown error")),
+            _ => Ok(DSError::UnknownError),
```
Maar-io, line 109: I think this should remain `DispatchError`, since the `try_from()` will fail and `execute_func()` will return `Result<RetVal, DispatchError>`

PierreOssun: `try_from()` will not fail. it will return `DSError::UnknownError` and will be decoded on contract side.
When your try to map DispatchError to DSError - it will map to an DSError in any case - so you can handle it in contract side. 
For example, if there is a new Error added in dapps-staking pallet that was not added to this TryForm impl - it will return UnknownError and we should update our DSError.
Still `execute_func()` can return an DispatchError if it fails on substrate level fucntions (ex: env().write) which will no be handled by the contract. Because query calls should not fail (only return a value not a Result). 
To Sum up:
- Any business error returned from dapps-staking pallet should be handled
- If the Err is not a business logic Err coming from dapps-staking it returns a Dispatch Error
- In a case where a business logic Err is returned from dapps-staking pallet is not mapped to DSError correctly it fails under UnknownError and should be immediatly added to DSError (this case should not happen actually as DSError should have exhaustive enum fiields of dapps-staking pallets errors - but we still cover it)

Maar-io: ok, I see, It will just be harder to debug in some cases since execute_func() will never fail.

Maar-io (approved):
LGTM

chain-extensions/types/dapps-staking/src/lib.rs
PierreOssun: It will be easy to debug because it return a RetVal.
It can still send Err at every line where there is an `?` but it is Err outside of dApps-staking logic.
`try_from()` only return DsError not DispatchError anymore

PR #90, opened 2022-09-06
Remove preapproval remnants

Dinonard:
**Pull Request Summary**

Removed deprecated pre-approval items from dapps-staking pallet and removed storage cleanup code.

Once Astar has been upgraded, we should ensure we update to latest `astar-frame` dependencies.

### DIFF
--- a/frame/dapps-staking/Cargo.toml
+++ b/frame/dapps-staking/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-dapps-staking"
-version = "3.6.2"
+version = "3.6.3"
 authors = ["Stake Technologies <devops@stake.co.jp>"]
 edition = "2021"
 homepage = "https://astar.network/"
--- a/frame/dapps-staking/src/benchmarking.rs
+++ b/frame/dapps-staking/src/benchmarking.rs
@@ -21,13 +21,11 @@ fn initialize<T: Config>() {
     let _ = Ledger::<T>::clear(u32::MAX, None);
     let _ = RegisteredDevelopers::<T>::clear(u32::MAX, None);
     let _ = RegisteredDapps::<T>::clear(u32::MAX, None);
-    let _ = PreApprovedDevelopers::<T>::clear(u32::MAX, None);
     let _ = GeneralEraInfo::<T>::clear(u32::MAX, None);
     let _ = ContractEraStake::<T>::clear(u32::MAX, None);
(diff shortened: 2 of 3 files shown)


Dinonard:
@shunsukew please check & approve.

Usually if upgrade involves bumping up pallet version in storage, there should be a check at the beginning which will skip it if the upgrade was already done. Right now this is causing problems when I try to use `try-runtime` because this check will fail (see the code).

Usually we (Astar) include runtime upgrades as part of runtime directly. E.g. as [here](https://github.com/AstarNetwork/Astar/commit/b97560442653291f3ef03bffc11b0996a6b2c8c2#diff-c8f1131d6e9583df89b2194443dd58b6766e82a3005a8dd7562f59f224585b81R1108). That way it can easily be removed.

Both approaches are fine, in a sense, but there should be some guards so upgrade isn't executed twice if it's done directly in the pallet.

akru (approved):

shunsukew (approved):

shunsukew:
> @shunsukew please check & approve.
> 
> Usually if upgrade involves bumping up pallet version in storage, there should be a check at the beginning which will skip it if the upgrade was already done. Right now this is causing problems when I try to use `try-runtime` because this check will fail (see the code).
> 
> Usually we (Astar) include runtime upgrades as part of runtime directly. E.g. as [here](https://github.com/AstarNetwork/Astar/commit/b97560442653291f3ef03bffc11b0996a6b2c8c2#diff-c8f1131d6e9583df89b2194443dd58b6766e82a3005a8dd7562f59f224585b81R1108). That way it can easily be removed.
> 
> Both approaches are fine, in a sense, but there should be some guards so upgrade isn't executed twice if it's done directly in the pallet.

I see. Thank you so much for teaching me.
If runtime upgrade logic is included in frame side and managed properly with versioning, then I can imagine some external chains which enable astar-frame pallet (just in case) can benefit from it. If I were a such user, I would feel very grateful with it at least.
I'll make sure to add some guards (checking upgrade already done or not. checking versioning in pallet storage for e.g?) in case with writing runtime upgrade in frame side. (I don't say it should necessarily be written in frame side though)

Dinonard:
@shunsukew that's correct, if this was a pallet which was used by other projects, we'd need to manage upgrade versions and keep the code for some time so everyone has had the chance to upgrade.

Since we know which runtimes use this pallet though, we can usually delete the code quite fast.

PR #92, opened 2022-09-07
XVM MVP v2

akru:
**Pull Request Summary**

This PR containts XVM v2 prototype.

Feature list:
 - [x] Contracts arguments encoding;
 - [ ] Error handling;
 - [ ] Return data handling;
 - [ ] XVM runtime interface & RPC.

**Check list**
- [ ] added unit tests
- [ ] updated documentation


chain-extensions/xvm/src/lib.rs
```diff
@@ -52,27 +51,19 @@ where
 
         match func_id {
             XvmFuncId::XvmCall => {
-                // TODO: correct weight calculation directly from pallet!
-                let weight = Weight::from_ref_time(1_000_000_000);
-                env.charge_weight(weight)?;
+                // We need to immediately charge for the worst case scenario. Gas equals Weight in pallet-contracts context.
```
Dinonard, line 54: I was confused at first by this because it seems so familiar lol :sweat_smile:

frame/collator-selection/src/tests.rs
```diff
@@ -117,7 +117,9 @@ fn cannot_register_candidate_if_too_many() {
 
         // reset desired candidates:
         <crate::DesiredCandidates<Test>>::put(1);
-        assert_ok!(CollatorSelection::register_as_candidate(RuntimeOrigin::signed(4)));
+        assert_ok!(CollatorSelection::register_as_candidate(
```
Dinonard, line 120: Why are unrelated files changed by the formatter? Are you using different settings?

frame/pallet-xvm/src/evm.rs
```diff
@@ -1,61 +1,67 @@
...
             gas_limit,
             max_fee_per_gas,
             None,
             None,
             Vec::new(),
-        );
+        )
+        .map_err(|_e| XvmCallError {
+            error: XvmError::ExecutionError(Vec::default()), // TODO: make error mapping make more sense
+            consumed_weight: 42u64, // TODO: res.actual_weight.map(|x| x.ref_time()).unwrap_or(context.max_weight),
```
Dinonard: We should go with worst case scenario here and specify `consumed_weight: context.max_weight`

frame/pallet-xvm/src/evm.rs
```diff
@@ -1,61 +1,67 @@
...
             target: "xvm::EVM::xvm_call",
             "EVM XVM call result: {:?}", res
         );
 
-        // TODO: return error if call failure
-        // TODO: return value in case of constant / view call
-        Ok(Default::default())
+        Ok(XvmCallOk {
+            output: Default::default(), // TODO: Fill output vec with response from the call
+            consumed_weight: 42u64, // TODO: res.actual_weight.map(|x| x.ref_time()).unwrap_or(context.max_weight),
```
Dinonard, line 71: You should extract correct `consumed_weight` from the `res`.

frame/pallet-xvm/src/wasm.rs
```diff
@@ -2,61 +2,64 @@
...
-            Weight::from_ref_time(gas_limit),
+            gas_limit.into(),
             None,
-            data,
-        );
+            input,
+        )
+        .map_err(|_e| XvmCallError {
+            error: XvmError::ExecutionError(Vec::default()), // TODO: make error mapping make more sense
+            consumed_weight: 42u64, //TODO: e.post_info.actual_weight.ref_time().unwrap_or(gas_limit),
```
Dinonard: Same comment as for EVM counter-part, we should use `context.max_weight` here.

frame/pallet-xvm/src/wasm.rs
```diff
@@ -2,61 +2,64 @@
...
             target: "xvm::WASM::xvm_call",
             "WASM XVM call result: {:?}", res
         );
 
-        // TODO: return error if call failure
-        // TODO: return value in case of constant / view call
-        Ok(Default::default())
+        Ok(XvmCallOk {
+            output: Default::default(), // TODO: Fill in with output from the call
+            consumed_weight: 42u64, //TODO: e.post_info.actual_weight.ref_time().unwrap_or(gas_limit),
```
Dinonard: Correct consumed weight can be extracted from the `res`.

Dinonard (changes requested):

chain-extensions/xvm/src/lib.rs
akru: I believe gas charging should be rewritten for new weighs, including comments. Let’s do it in XVMv3

frame/pallet-xvm/src/evm.rs
akru: Yes, but it not works with new weighs, so, I just hack it to make it works.

frame/collator-selection/src/tests.rs
akru: I guess it’s new astar-frame formatting settings

frame/pallet-xvm/src/evm.rs
akru: Yes, but next time

frame/pallet-xvm/src/wasm.rs
akru: For sure

chain-extensions/xvm/src/lib.rs
```diff
@@ -52,27 +51,19 @@ where
...
-                    vm_id,
-                    to,
-                    input,
-                    metadata,
-                } = env.read_as_unbounded(env.in_len())?;
+                let XvmCallArgs { vm_id, to, input } = env.read_as_unbounded(env.in_len())?;
 
-                // TODO: rethink this? How do we get valid env in chain extension? Need to know which data to encode.
-                // gas limit, taking into account used gas so far?
                 let _origin_address = env.ext().address().clone();
```
PierreOssun, line 62: I think you can delete these 2 unused values.
`_origin_address` will only be used in a case of an enum where contract can chose to use caller() or address()
`_value` will be used only when calling payable function will work

contracts/xvm/Cargo.toml
```diff
@@ -9,20 +9,24 @@ repository = "https://github.com/AstarNetwork/astar-frame"
 description = "Chain extension for X-VM calls"
```
PierreOssun: Is this contract redundant? As we have ink! SDK

akru: I guess, yes

akru: Let me remove it in next PR

chain-extensions/xvm/src/lib.rs
Dinonard: I meant to say that the reason this code looked so familiar is that I was the author. In the PR that added weight handling. xD

But it was like 2 months ago so I've forgotten about it.

This handling should be fine - some stuff needs to be benchmarked but that won't change the current weights too much.

frame/pallet-xvm/src/evm.rs
Dinonard: Can you please explain more? What doesn't work?

frame/pallet-xvm/src/evm.rs
Dinonard: Why next time?
Weights should be handled properly in v2, right?

And it's a very minor fix. I guess same question as above, but why is the "correct" code commented out?

frame/pallet-xvm/src/evm.rs
akru: I believe problem here is different types for `Weight` and context types. We should rework it using new substrate weight system.

frame/pallet-xvm/src/evm.rs
akru: Unfortunately this code no more works correctly, it assumes that weight is a number, but it is struct now.
I just want merge this PR because it takes so long time and proper weighting isn’t the main aim of this PR.

Dinonard: I see - however, the struct that was introduced was simply a wrapper around the old u64 value.
That was the initial step, the next one being adding the PoV size.

For this particular case, in the PR, all you need to do is this:

```
consumed_weight: res.actual_weight.map(|x| x.ref_time()).unwrap_or(context.max_weight.ref_time())
```
Just add `.ref_time()` to the `max_weight`.

chain-extensions/xvm/src/lib.rs
akru: Fixed

frame/pallet-xvm/src/evm.rs
akru: Not that simple but fixed

frame/pallet-xvm/src/wasm.rs
akru: Fixed

frame/pallet-xvm/src/wasm.rs
akru: Fixed

contracts/xvm/Cargo.toml
akru: Removed

akru: Removed

Dinonard (approved):
LGTM!

PR #94, opened 2022-09-13
XVM gas handling prototype

Dinonard:
**Pull Request Summary**

The purpose of this PR is to provide a prototype implementation for gas handling in XVM.

### Description

When executing a contract within the context of XVM, different VMs and gas handling schemas will be present.
E.g. the most basic example is EVM and Partity's WASM - gas vs. native weight. Each unit of these measures represents
a different concept, although similar. However, both are encapsulated by **substrate's** native way of handling weight (or gas).

So it's natural to keep the handling of *gas* in XVM  context tied to the native weight system.

### Current Solution

We rely on native weight handling for XVM *gas* handling.

To take an example when EVM contract calls WASM contract:
- EVM contract call is made by a user who sets **gas_limit**
- during EVM contract execution, some gas is consumed and a XVM call is made towards a WASM contract
- XVM precompile handles the call - it reads the remaining available gas, converts it to native *weight* and passes it on as `max_weight` via an XVM call
- XVM pallet detects which VM is being called and delegates call handling to it, also passing the `max_weight` value to WASM handler
- WASM handler executes the call, limiting it by `max_weight`, and in the return value returns how much weight was actually spent
- weight of the XVM call is updated (it can only be REDUCED compared to the initial setting)
- EVM precompile handler detects the real weight used for the XVM call, converts it to gas and refunds unused gas if needed

Same goes for `WASM --> EVM` direction.

### Open Issues, Questions & Suggestions

1. We should remove Async VM until we actually need to implement it.
2. Using wrapped `Call` isn't good enough if we want to have a return value..
3. Weight in the XVM call should be subtracted by XVM execution cost itself. Requires proper benchmarking.
5. We should limit XVM call depth. Probably a good starting value is 1 or 2.
6. XvmContext needs to be revised - imo, contract shouldn't care about limiting gas consumption - it's implicit from the call. I'm not even sure what exactly should be passed via it.

**Check list**
- [x] Receive feedback from team (1)
- [x] Apply first feedback
- [x] TBD

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,5 +21,5 @@ members = [
 	"precompiles/xcm",
 	"precompiles/xvm",
 
-	"contracts/*",
+	# "contracts/*",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,5 +21,7 @@ members = [
 	"precompiles/xcm",
 	"precompiles/xvm",
 
+	# Current build fails due to error associated with incorrect nightly usage
+	# but we use one from polkadot release so we probably shouldn't change it yet.
 	# "contracts/*",
 ]
--- a/frame/pallet-xvm/src/lib.rs
+++ b/frame/pallet-xvm/src/lib.rs
@@ -14,6 +14,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use codec::{Decode, Encode};
+use frame_support::weights::Weight;
 use sp_runtime::{traits::Member, RuntimeDebug};
 use sp_std::prelude::*;
 
@@ -42,6 +43,13 @@ pub enum XvmError {
     ExecutionError(Vec<u8>),
 }
 
+// TODO: Currently our precompile/chain-extension calls rely on direcy `Call` usage of XVM pallet.
+// This is perfectly fine when we're just calling a function in other VM and are interested whether the call was
+// successful or not.
+//
+// Problem arises IF we want to get back arbitrary read value from the other VM - `DispatchResultWithPostInfo` isn't enough for this.
+// We need to receive back a concrete value back from the other VM.
+
 #[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, scale_info::TypeInfo)]
 pub struct XvmCallOk {
     /// Output of XVM call. E.g. if call was a query, this will contain query response.
@@ -61,11 +69,20 @@ pub struct XvmCallError {
 
 pub type XvmResult = Result<XvmCallOk, XvmCallError>;
 
+pub fn consumed_weight(result: &XvmResult) -> Weight {
+    match result {
+        Ok(res) => res.consumed_weight,
+        Err(err) => err.consumed_weight,
+    }
+}
+
 /// XVM context consist of unique ID and optional execution arguments.
 #[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, scale_info::TypeInfo)]
 pub struct XvmContext {
     /// Identifier (should be unique for each VM in tuple).
     pub id: VmId,
+    /// Max allowed weight for the call
+    pub max_weight: Weight,
     /// Encoded VM execution environment.
     pub env: Option<Vec<u8>>,
 }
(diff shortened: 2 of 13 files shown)


frame/pallet-xvm/src/lib.rs
```diff
@@ -29,11 +30,62 @@ pub mod evm;
...
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, scale_info::TypeInfo)]
+pub enum XvmError {
+    VmNotRecognized,
+    EncodingFailure,
+    ContextConversionFailed,
+    OutOfGas,
+    ExecutionError(Vec<u8>),
+}
+
+// TODO: Currently our precompile/chain-extension calls rely on direcy `Call` usage of XVM pallet.
```
0x7CFE: `s/direcy/direct/`

frame/pallet-xvm/src/lib.rs
```diff
@@ -29,11 +30,62 @@ pub mod evm;
...
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, scale_info::TypeInfo)]
+pub struct XvmCallError {
+    /// Result of XVM call
+    // TODO: use XvmError enum from pallet? Perhaps that's a better approach. Or at least provide mapping?
+    error: XvmError,
+    /// Total consumed weight. This is in context of Substrate (1 unit of weight ~ 1 ps of execution time)
+    consumed_weight: u64,
+}
+
+pub type XvmResult = Result<XvmCallOk, XvmCallError>;
```
0x7CFE, line 76: pub type should probably have a rustdoc

frame/pallet-xvm/src/lib.rs
```diff
@@ -166,7 +229,7 @@ pub trait XvmCodec {
     fn convert(input: Vec<u8>, metadata: Vec<u8>) -> Result<Vec<u8>, Vec<u8>>;
 }
 
-/// Identity codec implementation: just pass input as is ignoring metadata.
+/// Identity codec implementation: ignoring metadata,
```
0x7CFE: mistyped comma?

frame/pallet-xvm/src/lib.rs
Dinonard: Added, although it's self-explanatory in a way.

Cargo.toml
```diff
@@ -21,5 +21,5 @@ members = [
 	"precompiles/xcm",
 	"precompiles/xvm",
 
-	"contracts/*",
+	# "contracts/*",
```
0x7CFE, line 26: Unintentionally left edit?

frame/pallet-xvm/src/pallet/mod.rs
```diff
@@ -68,21 +79,24 @@ pub mod pallet {
             message: Vec<u8>,
         ) -> DispatchResultWithPostInfo {
             let from = ensure_signed(origin)?;
-            let result = T::AsyncVM::xvm_send(context, from, to, message);
+            let _result = T::AsyncVM::xvm_send(context, from, to, message);
+
+            Self::deposit_event(Event::<T>::XvmSend {
+                result: Default::default(),
+            }); // TODO: change this
```
0x7CFE: Not very informative comment. At least write a couple of words on what need to be changed and why.

frame/pallet-xvm/src/pallet/mod.rs
```diff
@@ -46,18 +46,29 @@ pub mod pallet {
...
+            let mut context = context;
+            context.max_weight = context.max_weight - PLACEHOLDER_WEIGHT;
+
             let result = T::SyncVM::xvm_call(context, from, to, input);
 
-            Self::deposit_event(Event::<T>::XvmCall { result });
-            Ok(().into())
+            Self::deposit_event(Event::<T>::XvmCall {
+                result: Ok(Default::default()),
+            }); // TODO: this event should probably be changed
```
0x7CFE: Changed to what? :)

0x7CFE (commented):
more grumbles

Cargo.toml
Dinonard: Added a comment.

It seems that after latest updates, compilation with `cargo check` fails.

`cargo contract check` should work but it also fails for me with another error so I'll have to check that in more detail.

Perhaps this also indicates that we shouldn't keep contracts under root `.toml` file :thinking:

frame/pallet-xvm/src/pallet/mod.rs
Dinonard: Agree, but it wasn't ready for review yet :D

akru (approved):

PR #95, opened 2022-09-13
Correct weight refund in ds-chain-extension

Dinonard:
**Pull Request Summary**

Removes `unwrap()` from dapps-staking chain-extension and replaces it with an exhaustive result check.


Maar-io (approved):
Thanks!

Issue #63, opened 2022-06-21
Re-bond Unbonding Chunks in dApps Staking


Dinonard:
# Description
Once stakers start the unbonding, their unbonded funds will remain locked during the `unbonding period`. During this period, these funds don’t generate any rewards neither can they be used for payment, transfer, etc.
This is normal and expected behavior.

However, we want to provide users an option to re-bond their unbonding chunks.
This allows them to continue earning rewards in dapps-staking and returns their unbonding chunks to locked state.

## Details
* this is only applicable for users who have unbonding chunks
* it doesn't matter whether chunks have completed the unbonding period or not
* when used, **all** chunks should be consumed and re-bonded
* user can only select one target contract on which they want to re-bond

## Required Tasks
- [ ] implement a new extrinsic call `rebond_chunks` (or something similar)
- [ ] unit tests
- [ ] runtime benchmarks code

## Estimated done.
<>

## Dependencies
This task has no dependencies but will require UI support to be utilized from the portal side.

PR #97, opened 2022-09-15
[Pending] Rebond unlocking chunks

shunsukew:
Solves https://github.com/AstarNetwork/astar-frame/issues/63

**Pull Request Summary**
Added feature `rebond_and_stake` unlocking chunks to contracts.

Once stakers start the unbonding, their unbonded funds will remain locked during the `unbonding period`. During this period, these funds neither generate any rewards nor can they be used for payment, transfer, etc.

One functionality we could provide is for stakers to re-bond their unbonding chunks. They could select unbonding chunks as a source when staking on a contract from UI, similar as they do for nomination_transfer. This will result in unbonding chunks being consumed and (re)staked on some contract.

**Check list**
- [x] added unit tests
- [x] added benchmark
- [x] added chain extension
- [x] added precompiles
- [ ] updated documentation


frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -570,6 +574,78 @@ pub mod pallet {
...
+        /// 
+        /// `value` must be more than the `minimum_balance` specified by `MinimumStakingAmount`
+        /// unless account already has bonded value equal or more than 'minimum_balance'.
+        /// 
+        /// The dispatch origin for this call must be _Signed_ by the staker's account.
+        #[pallet::weight(T::WeightInfo::rebond_and_stake())]
+        pub fn rebond_and_stake(
+            origin: OriginFor<T>,
+            contract_id: T::SmartContract,
+            #[pallet::compact] value: BalanceOf<T>,
```
Dinonard: In order to save you some time with dev & test work - this should be removed.
The call should be kept as simple as possible - no sorting and such stuff since it doesn't provide any real value to the user.

They can only rebond ALL chunks.

Dinonard (changes requested):

frame/dapps-staking/src/pallet/mod.rs
shunsukew: Thank you so much for notifying me in the early stage!

chain-extensions/dapps-staking/src/lib.rs
```diff
@@ -333,6 +335,28 @@ impl<T: pallet_dapps_staking::Config> ChainExtensionExec<T> for DappsStakingExte
                     Ok(_) => Ok(RetVal::Converging(DSError::Success as u32)),
                 };
             }
+
+            DappsStakingFunc::RebondAndStake => {
+                let contract_bytes: [u8; 32] = env.read_as()?;
```
0x7CFE: I have seen this `[u8; 32]` in several places already. If if's so common, then maybe we should assign a type for it? That way it would be easier to understand, what this is and what can be done with that.

chain-extensions/types/dapps-staking/src/lib.rs
```diff
@@ -65,6 +65,8 @@ pub enum DSError {
     NominationTransferToSameContract = 26,
     /// Unexpected reward destination value
     RewardDestinationValueOutOfBounds = 27,
+    /// There are no previously unbonded funds that can be reboneded and staked.
```
0x7CFE: Other comments do not end with trail `.`. Probably better to stick with the rest.

frame/dapps-staking/src/benchmarking.rs
```diff
@@ -217,6 +217,24 @@ benchmarks! {
...
+        let _ = T::Currency::make_free_balance_be(&staker, BalanceOf::<T>::max_value());
+        let stake_amount = BalanceOf::<T>::max_value() / 2u32.into();
+        let unstake_amount = stake_amount / 2u32.into();
+
+        DappsStaking::<T>::bond_and_stake(RawOrigin::Signed(staker.clone()).into(), contract_id.clone(), stake_amount)?;
+        DappsStaking::<T>::unbond_and_unstake(RawOrigin::Signed(staker.clone()).into(), contract_id.clone(), unstake_amount)?;
+    }: _(RawOrigin::Signed(staker.clone()), contract_id.clone())
+    verify {
+        assert_last_event::<T>(Event::<T>::RebondAndStake(staker, contract_id, unstake_amount).into());
+    }
```
0x7CFE: nit: I'd like to have some comments describing what is actually happening.

shunsukew:
compilation started to fail after pushing small commit https://github.com/AstarNetwork/astar-frame/pull/97/commits/852fb39ac8f512e7c7765a78e33be761e81dbc99. Why, will check

chain-extensions/dapps-staking/src/lib.rs
```diff
@@ -336,6 +338,28 @@ impl<T: pallet_dapps_staking::Config> ChainExtensionExec<T> for DappsStakingExte
...
+
+                let base_weight =
+                    <T as pallet_dapps_staking::Config>::WeightInfo::rebond_and_stake();
+                env.charge_weight(base_weight)?;
+
+                let caller = env.ext().address().clone();
+                let call_result = pallet_dapps_staking::Pallet::<T>::rebond_and_stake(
+                    RawOrigin::Signed(caller).into(),
+                    contract,
+                );
```
0x7CFE, line 354: Was kinda confused since `rebond_and_stake` is called twice in a row. I do understand that these are two different calls, but still this is a bit confusing. Maybe we should add couple of comments to reduce overall WTF/minute.

chain-extensions/types/dapps-staking/src/lib.rs
```diff
@@ -120,27 +123,29 @@ pub enum Contract<Account> {
     Wasm(Account),
 }
 
+pub type ContractBytes = [u8; 32];
```
0x7CFE, line 126: Since this type is in public it would be a good idea to add some explanatory comment. What is this type, why it 32 bytes long, what is the purpose of that type, etc.

0x7CFE (commented):
LGTM aside of small nits

chain-extensions/dapps-staking/src/lib.rs
shunsukew: This can be said for other functions too. it's a more universal topic.
I thought it's very clear they are different and what they do because they are different methods, one is `WeightInfo`'s and another is `Pallet`'s.

shunsukew:
@Dinonard I also added chain extension and precompiles, but I started to think they should be separately released (in a different PR) after this function become stable. What do you think?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -570,6 +574,64 @@ pub mod pallet {
             Ok(().into())
         }
 
+        /// Lock up and stake unbonded chunks of origin account.
+        ///
+        /// All unbonding chunks will be used and staked to the specified contract.
+        ///
+        /// The dispatch origin for this call must be _Signed_ by the staker's account.
```
Dinonard, line 581: Please also check top of lib.rs file and update comments there if needed.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -570,6 +574,64 @@ pub mod pallet {
...
+                value_to_stake,
+                current_era,
+            )?;
+
+            ledger.unbonding_info.unlocking_chunks = Vec::<UnlockingChunk<BalanceOf<T>>>::default();
+
+            GeneralEraInfo::<T>::mutate(&current_era, |value| {
+                if let Some(x) = value {
+                    x.staked = x.staked.saturating_add(value_to_stake);
+                    x.locked = x.locked.saturating_add(value_to_stake);
```
Dinonard, line 618: This is incorrect - TVL doesn't increase after this since unbonding chunks are still considered to be locked.

UT should be updated to catch this.

frame/dapps-staking/src/testing_utils.rs
```diff
@@ -353,6 +353,67 @@ pub(crate) fn assert_unbond_and_unstake(
...
+
+    // Verify the remaining states
+    assert_eq!(
+        final_state.era_info.staked,
+        init_state.era_info.staked + expected_stake_amount
+    );
+    assert_eq!(
+        final_state.era_info.locked,
+        init_state.era_info.locked + expected_stake_amount
+    );
```
Dinonard, line 406: This is incorrect, see my comment above.

frame/dapps-staking/src/tests.rs
```diff
@@ -1059,6 +1059,71 @@ fn unbond_and_unstake_with_no_chunks_allowed() {
...
+        let contract_id = MockSmartContract::Evm(H160::repeat_byte(0x01));
+        assert_register(10, &contract_id);
+
+        assert_noop!(
+            DappsStaking::rebond_and_stake(Origin::signed(staker_id), contract_id),
+            Error::<TestRuntime>::NothingToRebond,
+        );
+    })
+}
+
```
Dinonard, line 1126: I'd suggest also covering the error cases with InsufficientValue or TooManyEraStakeValues.

Not mandatory, just a suggestion.

Dinonard (commented):
Thanks for working on this!

About your precompile/chain-extension questions, I think we should keep them out of this PR. And maybe we shouldn't add them at all. 
Your comment about feature being stable makes sense and I'd also add the point that we'll be redesigning dapps-staking very soon. It's not necessarily true that this function will be compatible with new version so I'd wait until then with it. That way we reduce number of incompatible legacy functions.

shunsukew:
> Thanks for working on this!
> 
> About your precompile/chain-extension questions, I think we should keep them out of this PR. And maybe we shouldn't add them at all. Your comment about feature being stable makes sense and I'd also add the point that we'll be redesigning dapps-staking very soon. It's not necessarily true that this function will be compatible with new version so I'd wait until then with it. That way we reduce number of incompatible legacy functions.

That makes sense.
In that case, I'd make this PR pending until the redesign and its implementation finish. I'll resume this PR work after that.

Dinonard:
> > Thanks for working on this!
> > About your precompile/chain-extension questions, I think we should keep them out of this PR. And maybe we shouldn't add them at all. Your comment about feature being stable makes sense and I'd also add the point that we'll be redesigning dapps-staking very soon. It's not necessarily true that this function will be compatible with new version so I'd wait until then with it. That way we reduce number of incompatible legacy functions.
> 
> That makes sense. In that case, I'd make this PR pending until the redesign and its implementation finish. I'll resume this PR work after that.

I'm not sure how feasible is that, tbh. We don't know if this will even be applicable then.

I don't see an issue with this being merged now but we should avoid precompiles/chain-extensions. Also, not sure how much it makes sense to ask UI team to do this since it will most likely be changed soon.

Anyhow, `draft` status that you used is probably the way to go.

PR #102, opened 2022-09-23
XCM precompile - transact

Dinonard:
**Pull Request Summary**

Support for using `Transact` instruction remotely on other chains via EVM smart contracts.

This enables smart contracts to send an arbitrary call to any chain Astar/Shiden has an open channel with.
User needs to specify a couple of arguments:
- `destination` - parachain Id or relay chain
- `payment asset & amount` - to use on the destination chain
  - user should have enough asset on the destination chain
- `weight` - total execution weight in the destination chain - sum of 4 XCM instructions executions + weight required to execute the encoded call via `Transact`
- `call` - remote encoded call - this is what will be executed on the destination chain 

The template message being sent looks like this:
1. DescendOrigin
2. WithdrawAsset
3. BuyExecution
4. Transact

User doesn't have control over this, except for the aforementioned configurable parameters.

**Open Questions**
- **Q1** - destination chain needs to support `DescendOrigin` and converting e.g. `AccountId32` into appropriate local representation. Will this work with chains like Moonbeam where `AccountKey20` is the default account representation type?
- **A** - yes because we don't control direct mapping of our address in the destination chain but most likely a mapping achieved by hashing our multilocation with additional fixed data to derive new address
- **Q2** - should there be `Refund` and `DepositAsset` instructions? Or should user ensure they use the correct weight and payment amount? This DOES make the transaction more expensive so perhaps it's better to leave it for future improvement.

**Check list**
- [x] extend UTs or create an item to improve XCM precompile tests
- [x] added unit tests
- [x] updated documentation

### DIFF
--- a/precompiles/xcm/src/lib.rs
+++ b/precompiles/xcm/src/lib.rs
@@ -27,7 +27,7 @@ mod tests;
 pub enum Action {
     AssetsWithdrawNative = "assets_withdraw(address[],uint256[],bytes32,bool,uint256,uint256)",
     AssetsWithdrawEvm = "assets_withdraw(address[],uint256[],address,bool,uint256,uint256)",
-    RemoteTransact = "remote_transact(uint256,bool,address,uint256,uint64,bytes,uint64)",
+    RemoteTransact = "remote_transact(uint256,bool,address,uint256,bytes,uint64)",
 }
 
 /// A precompile that expose XCM related functions.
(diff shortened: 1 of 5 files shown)


precompiles/xcm/src/lib.rs
```diff
@@ -163,4 +165,79 @@ where
...
+            .map_err(|_| revert("Failed to reanchor fee asset"))?;
+
+        // Prepare XCM
+        let xcm = Xcm(vec![
+            WithdrawAsset(fee_multilocation.clone().into()),
+            BuyExecution {
+                fees: fee_multilocation.clone().into(),
+                weight_limit: WeightLimit::Unlimited,
+            },
+            Transact {
```
akru, line 222: So, any call of this precompile will be executed from parachain account?

precompiles/xcm/src/lib.rs
```diff
@@ -163,4 +165,79 @@ where
...
+        let fee_amount = input.read::<U256>()?;
+
+        let remote_call: Vec<u8> = input.read::<Bytes>()?.into();
+        let transact_weight = input.read::<u64>()?;
+
+        log::trace!(target: "xcm-precompile:remote_transact", "Raw arguments: para_id: {}, is_relay: {}, fee_asset_addr: {:?}, \
+         fee_amount: {:?}, remote_call: {:?}, transact_weight: {}",
+        para_id, is_relay, fee_asset_addr, fee_amount, remote_call, transact_weight);
+
+        // Process arguments
```
0x7CFE, line 187: The comment actually says nothing. Probably OK to remove or rephrase if that's important.

0x7CFE (approved):
LGTM

precompiles/xcm/src/lib.rs
Dinonard: I don't like these big methods so I put it here since I'm past raw arguments now. Guess it's a personal preference.

precompiles/xcm/src/lib.rs
Dinonard: I was confused about this, and had discussion both with Crust & Moonbeam team.
Ended up with a stack-exchange post [here](https://substrate.stackexchange.com/questions/5066/originkind-in-xcms-transact).

Dinonard: @akru I hadn't noticed your question until today. I will merge soon unless there's some other comment.

PR #106, opened 2022-10-14
dApps staking max encoded len

Dinonard:
**Pull Request Summary**

Removes `without_storage_info` from the `pallet-dapps-staking` crate.
Introduces derivation of `MaxEncodedLen` trait to all possible structs, enums and types.

For the `UnbondingInfo` and `StakerInfo`, an assumption is introduced that their length won't go over a specific constant.
Since we're fully in control of runtime code, we can be fairly sure this limitation won't be exceeded. Also, dApps staking is a stable feature and there is no reason to modify current parameters from what they currently are.

See code more more detailed explanation & comments.

### DIFF
--- a/frame/dapps-staking/src/lib.rs
+++ b/frame/dapps-staking/src/lib.rs
@@ -227,7 +227,7 @@ impl<Balance: AtLeast32BitUnsigned + Copy + MaxEncodedLen> EraStake<Balance> {
 ///
 /// **NOTE:** It is important to understand that staker **DID NOT** claim any rewards during this period.
 ///
-#[derive(Encode, Decode, Clone, Default, PartialEq, Eq, RuntimeDebug, TypeInfo)] // TODO! MaxEncodedLen
+#[derive(Encode, Decode, Clone, Default, PartialEq, Eq, RuntimeDebug, TypeInfo)]
 pub struct StakerInfo<Balance: AtLeast32BitUnsigned + Copy + MaxEncodedLen> {
     // Size of this list would be limited by a configurable constant
     stakes: Vec<EraStake<Balance>>,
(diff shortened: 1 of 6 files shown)


frame/dapps-staking/src/lib.rs
```diff
@@ -80,8 +80,22 @@ pub type BalanceOf<T> =
...
+// To fix this 100% correctly, we'd need to do one of the following:
+//
+// - Use `BoundedVec` instead of `Vec` and do storage migration
+// - Introduce a new type `S: Get<u32>` into the aforementioned structs and use it to inject max allowed size, thus allowing us to correctly calculate max encoded len
+//
+// The issue with first approach is that it requires storage migration which we want to avoid unless it's really necessary.
+// The issue with second approach is that it makes code much more difficult to work with since all of it will be ridden with injections of the `S` type.
+//
+// Since dApps staking has been stable for long time and there are plans to redesign & refactor it, doing neither of the above makes sense, timewise.
+// So we use an assumption that Vec length won't go over the following constant.
```
0x7CFE: Do we have max line length specified somewhere in our code guidelines? I believe the format check should handle this.

frame/dapps-staking/src/tests.rs
```diff
@@ -2132,3 +2132,12 @@ pub fn set_contract_stake_info() {
         );
     })
 }
+
+#[test]
+fn custom_max_encoded_len() {
+    // 10 x (4 + 16) + 1 = 10 x 20 + 1 = 201
+    assert_eq!(UnbondingInfo::<u128>::max_encoded_len(), 201);
```
0x7CFE: Why not write that comment as an expression instead of the constant?

frame/dapps-staking/src/lib.rs
Dinonard: Not sure whether that's specified anywhere.

Format checker can indeed check this but we're using default settings for formatting.
- https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#comment_width
- https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#wrap_comments

Comment wrapping is disabled by default.

frame/dapps-staking/src/tests.rs
Dinonard: Sure, that sounds good.

0x7CFE (approved):

PR #108, opened 2022-10-19
Add workaround to reserve transfer native assets

0x7CFE:
Currently we identify assets being transferred by specifying an H160 style address (XC20), so this prevents us from sending native token since there's no representation for it.

This PR adds a workaround to interpret address `0x0` as native token to be sent (`parents: 0, interior: Here`), it can be used to pay for transaction.


precompiles/xcm/src/lib.rs
```diff
@@ -280,6 +280,18 @@ where
...
+        let value = handle.context().apparent_value;
+        if value > U256::zero() {
+            assets.push(
+                MultiLocation {
+                    parents: 0,
+                    interior: Here,
+                }
+            );
+
+            amounts.push(value.low_u128());
```
Dinonard: This is exploitable and unsafe - please do the check as on line L271.

Dinonard (commented):
Please check this part: `handle.check_function_modifier(FunctionModifier::NonPayable)?;` in lib.rs.

It should be modified to decide `FunctionModifier` based on the call handle.

Dinonard:
At this point, it might be worth doing the UT improvement item for XCM precompiles.

It would be good to check that in case `msg.value > 0` we get a modified asset list in the dispatched XCM.

0x7CFE:
> It would be good to check that in case `msg.value > 0` we get a modified asset list in the dispatched XCM.

Agree. Current solution looks ad-hoc and fragile.

precompiles/xcm/src/lib.rs
```diff
@@ -280,6 +291,21 @@ where
             return Err(revert("Assets resolution failure."));
         }
 
+        let value = handle.context().apparent_value;
+        if value > U256::zero() {
+            handle.check_function_modifier(FunctionModifier::NonPayable)?;
```
Dinonard: Seems redundant if you've already ensured these checks are made in the `execute` function.

Dinonard (commented):
Do you need to declare functiions in the `.sol` file as payable?
Or can you test in Remix without it?

precompiles/xcm/src/lib.rs
0x7CFE: Indeed.

akru: I believe function should be `payable` to get some `value` onboard.

Dinonard (approved):
LGTM!

Please also bump the version in the `.toml` file.

akru (approved):

precompiles/xcm/src/lib.rs
```diff
@@ -54,10 +54,21 @@ where
 
         let selector = handle.read_selector()?;
 
-        handle.check_function_modifier(FunctionModifier::NonPayable)?;
+        // Check function modifiers
+        match selector {
+            Action::AssetsReserveTransferNative | Action::AssetsReserveTransferEvm => {
+                // these precompiles use EVM msg.value to deal with native assets,
+                // this require function to be defined as payable
+                handle.check_function_modifier(FunctionModifier::Payable)?;
```
0x7CFE: Obsolete. Will remove

precompiles/xcm/src/lib.rs
```diff
@@ -266,7 +277,17 @@ where
             .iter()
             .cloned()
             .filter_map(|address| {
-                R::address_to_asset_id(address.into()).and_then(|x| C::reverse_ref(x).ok())
+                let address: H160 = address.into();
+
+                // Special case for native asset
+                if address == H160::from([0xff; 20]) {
```
Dinonard: `H160::MAX` is already used for relay chain native asset.

You should use `MAX -1`. Or some other perhaps, so they don't just differ by 1 bit.

Dinonard (changes requested):

precompiles/xcm/src/lib.rs
0x7CFE: done

0x7CFE:
Documented recent changes in API in https://github.com/AstarNetwork/astar-docs/pull/130

precompiles/xcm/src/lib.rs
```diff
@@ -279,9 +268,17 @@ where
...
+                let marker_address = H160::from([
+                    0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                    0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE,
+                ]);
+
                 // Special case for native asset
-                if address == H160::from([0xff; 20]) {
-                     Some(MultiLocation {
+                if address == marker_address {
+                    Some(MultiLocation {
```
Dinonard: Just a suggestion: `Here.into()`

Dinonard (changes requested):
Please add unit tests, and cover the new case.

precompiles/xcm/src/tests.rs
```diff
@@ -148,33 +150,17 @@ fn reserve_transfer_works() {
 
 #[test]
 fn reserve_transfer_native_works() {
```
Dinonard: Still good to test function with byte32 argument - please return it.

precompiles/xcm/src/tests.rs
```diff
@@ -183,4 +169,31 @@ fn reserve_transfer_native_works() {
...
+    let (location, xcm) = take_sent_xcm().pop().unwrap();
+    assert_eq!(
+        location,
+        MultiLocation {
+            parents: 1,
+            interior: Here
+        }
+    );
+
+    println!("XCM: {:#?}", xcm);
```
Dinonard: Can this be cleaned up?

precompiles/xcm/src/tests.rs
```diff
@@ -145,3 +147,53 @@ fn reserve_transfer_works() {
...
+
+    assert_matches!(
```
Dinonard: Great!

Please also include tests for non-native assets.

precompiles/xcm/src/tests.rs
0x7CFE: Is it bad? It would not show during normal testing. But if a test fails it would show the actual XCM as well as a failed assertion.

akru (approved):

precompiles/xcm/src/tests.rs
Dinonard: Depends how you look at it - if you're debugging an UT, you might want to add custom traces yourself where you need them. I assume this was added for specific reasons, due to barrier and asset transactor blocking the outgoing call.

Anyhow, your call.

precompiles/xcm/src/tests.rs
0x7CFE: Done

precompiles/xcm/src/tests.rs
0x7CFE: Removed for the sake of consistency with other tests

precompiles/xcm/src/tests.rs
```diff
@@ -108,7 +110,7 @@ fn remote_transact_works() {
 }
 
 #[test]
-fn reserve_transfer_works() {
+fn reserve_transfer_non_native_works() {
```
Dinonard: Perhaps a bit misleading - it's still native but not native currency. All of these assets are in fact _native_ and our parachain acts as reserve for them.

precompiles/xcm/src/tests.rs
```diff
@@ -144,4 +146,119 @@ fn reserve_transfer_works() {
...
+                }
+            ]
+
+            if fees.contains(&non_native_asset) && assets.contains(&non_native_asset)
+        );
+    }
+}
+
+#[test]
+fn reserve_transfer_native_works() {
```
Dinonard: Similar comment to above - perhaps `native_currency`?

Dinonard (commented):
Just a suggestion.

precompiles/xcm/src/tests.rs
0x7CFE: Feel free to suggest better name. Naming is hard.

Dinonard (approved):
LGTM :+1:

PR #109, opened 2022-10-21
Remove chain-extension-trait package

akru:
**Pull Request Summary**

The purpose of this PR is 
* remove cross-links between CE implementations trough `chain-extension-trait` package;
* optimize CE definition on runtime level (linked [PR](https://github.com/AstarNetwork/Astar/pull/768))

The problem is when we trying to development new CE usually we doing following steps:
1. fork development branch;
2. add dependency with new package pointed into development branch.

But, linking all CE to same `Trait` package makes requirement for all CE use it, *and use it with same `source`*, that means **touch one CE requires repoint all CEs to development branch**. It looks inconvenient in terms of independent development for may CEs. The purpose of this PR is making CE development more independent.

**Dapps Staking CE 1.1**

This PR aslo includes some code optimizations for dapps-staking CE. `Contract` type was removed and used runtime-level converters from `[u8; 32]` to `SmartContract` types. It looks much more safe rather than hacks with codec, runtime level converter is trivial and looks much more like trait-driven programming.

So, instead of byte manipulations hacks we just make requirement to have `From<[u8; 32]>` trait impl somewhere (usually where `SmartContract` type defined). 

**Check list**
- [ ] added unit tests
- [ ] updated documentation


chain-extensions/dapps-staking/src/lib.rs
```diff
@@ -65,19 +64,27 @@ impl TryFrom<u32> for DappsStakingFunc {
...
+impl<T> Default for DappsStakingExtension<T> {
+    fn default() -> Self {
+        DappsStakingExtension(PhantomData)
+    }
+}
+
+impl<T> ChainExtension<T> for DappsStakingExtension<T>
+where
+    T: pallet_dapps_staking::Config + pallet_contracts::Config,
+    <T as pallet_dapps_staking::Config>::SmartContract: From<[u8; 32]>,
```
Dinonard, line 79: I agree this looks more elegant but I'm not fond of assuming that this will be the only `256` bit value (even though it probably will be). Seems just a little bit hacky.

Dinonard (approved):
LGTM

Nice PR description, very easy to understand what you want to accomplish! :+1:

chain-extensions/dapps-staking/src/lib.rs
akru: Technically we can try to release it to `SysConfig::AccountId` type, let me try

chain-extensions/xvm/src/lib.rs
```diff
@@ -29,21 +29,25 @@ impl TryFrom<u32> for XvmFuncId {
...
 }
 
-pub struct XvmExtension;
+/// XVM chain extension.
+pub struct XvmExtension<T>(PhantomData<T>);
 
-impl<T> ChainExtensionExec<T> for XvmExtension
+impl<T> Default for DappsStakingExtension<T> {
+    fn default() -> Self {
+        RmrkExtension(PhantomData)
```
Maar-io, line 37: RMRK?

chain-extensions/xvm/src/lib.rs
```diff
@@ -29,21 +29,25 @@ impl TryFrom<u32> for XvmFuncId {
     }
 }
 
-pub struct XvmExtension;
+/// XVM chain extension.
+pub struct XvmExtension<T>(PhantomData<T>);
 
-impl<T> ChainExtensionExec<T> for XvmExtension
+impl<T> Default for DappsStakingExtension<T> {
```
Maar-io: DappsStaking here?

chain-extensions/xvm/src/lib.rs
Maar-io: RMRK? perhaps copy/paste error

Maar-io (changes requested):

chain-extensions/rmrk/src/lib.rs
```diff
@@ -18,23 +17,27 @@ use rmrk_traits::{
...
 
-pub struct RmrkExtension<R>(PhantomData<R>);
+/// RMRK chain extension.
+pub struct RmrkExtension<T>(PhantomData<T>);
 
-impl<
-        T: pallet_rmrk_core::Config
-            + pallet_uniques::Config<CollectionId = CollectionId, ItemId = NftId>,
-    > ChainExtensionExec<T> for RmrkExtension<T>
+impl<T> Default for DappsStakingExtension<T> {
```
Maar-io: copy/paste?

chain-extensions/rmrk/src/lib.rs
```diff
@@ -2,13 +2,12 @@
```
Maar-io, line 2: in this file, please change `unwrap()` to `unwrap_or_default()`

Maar-io (changes requested):

Maar-io (approved):

PR #112, opened 2022-11-08
DappsStaking bench improvements

shunsukew:
**Pull Request Summary**
Dapps Staking `CurrentEra` storage value is accessed by `on_initialize` hook at every block.
That means current era value is available in overlay cache, thus we can ignore storage access from benchmarking.

New attribute macro `#[pallet::whitelist_storage]` registers Storage items to whitelists (which will be ignored in benchmarking), see more details https://github.com/paritytech/substrate/pull/12205.

This requires Astar repo updates too, but adding attribute can be done independently.

This PR fixes format issue as well `cargo fmt`, `taplo fmt` to pass merge blocking.

### DIFF
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -103,6 +103,7 @@ pub mod pallet {
 
     /// Denotes whether pallet is disabled (in maintenance mode) or not
     #[pallet::storage]
+    #[pallet::whitelist_storage]
     #[pallet::getter(fn pallet_disabled)]
     pub type PalletDisabled<T: Config> = StorageValue<_, bool, ValueQuery>;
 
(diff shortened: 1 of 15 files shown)


frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -114,6 +114,7 @@ pub mod pallet {
 
     /// The current era index.
     #[pallet::storage]
+    #[pallet::whitelist_storage]
```
Dinonard, line 118: Since we're doing this, please whitelist all necessary storage:
* PalletDisabled
* ForceEra
* NextEraStartingBlock

Dinonard (changes requested):

frame/dapps-staking/src/pallet/mod.rs
shunsukew: That's true. I updated. thank you for pointing out

Dinonard (approved):
Great!

Please also bump the minor version in `.toml` file before merging.

PR #113, opened 2022-11-10
Beneficiary

kaichaosun:
**Pull Request Summary**

**Check list**
- [x] added unit tests
- [x] updated documentation

### DIFF
--- a/frame/dapps-staking/src/pallet/mod.rs
+++ b/frame/dapps-staking/src/pallet/mod.rs
@@ -180,6 +180,19 @@ pub mod pallet {
         ValueQuery,
     >;
 
+    /// Beneficiary of staking rewards on perticular contract.
+    /// `(staker, contract_id) -> beneficiary_account_id`
+    #[pallet::storage]
+    #[pallet::getter(fn staking_beneficiary)]
+    pub type RewardsBeneficiary<T: Config> = StorageDoubleMap<
+        _,
+        Twox64Concat,
+        T::AccountId,
+        Blake2_128Concat,
+        T::SmartContract,
+        T::AccountId,
+    >;
+
     /// Stores the current pallet storage version.
     #[pallet::storage]
     #[pallet::getter(fn storage_version)]
@@ -233,6 +246,12 @@ pub mod pallet {
             BalanceOf<T>,
             T::SmartContract,
         ),
+        /// Staking rewards beneficiary is set
+        BeneficiarySet(T::AccountId, T::SmartContract, T::AccountId),
+        /// Staking rewards beneficiary is removed
+        BeneficiaryRemoved(T::AccountId, T::SmartContract),
+        /// Staking rewards beneficiary is updated
+        BeneficiaryUpdated(T::AccountId, T::SmartContract, T::AccountId),
     }
 
     #[pallet::error]
@@ -291,6 +310,10 @@ pub mod pallet {
         NotActiveStaker,
         /// Transfering nomination to the same contract
         NominationTransferToSameContract,
+        /// There is no beneficiary set for the staker per contract
+        BeneficiaryNotSet,
+        /// Beneficiary used is not valid
+        InvalidBeneficiary,
     }
 
     #[pallet::hooks]
@@ -751,7 +774,9 @@ pub mod pallet {
                 staker_info.latest_staked_value(),
             );
 
-            if should_restake_reward {
+            let beneficiary = RewardsBeneficiary::<T>::get(&staker, &contract_id).unwrap_or(staker.clone());
+
+            if staker == beneficiary && should_restake_reward {
                 staker_info
                     .stake(current_era, staker_reward)
                     .map_err(|_| Error::<T>::UnexpectedStakeInfoEra)?;
@@ -797,9 +822,9 @@ pub mod pallet {
                 ));
             }
 
-            T::Currency::resolve_creating(&staker, reward_imbalance);
+            T::Currency::resolve_creating(&beneficiary, reward_imbalance);
             Self::update_staker_info(&staker, &contract_id, staker_info);
-            Self::deposit_event(Event::<T>::Reward(staker, contract_id, era, staker_reward));
+            Self::deposit_event(Event::<T>::Reward(beneficiary, contract_id, era, staker_reward));
 
             Ok(Some(if should_restake_reward {
                 T::WeightInfo::claim_staker_with_restake()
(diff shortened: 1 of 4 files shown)


frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -961,6 +987,79 @@ pub mod pallet {
...
+
+            Self::deposit_event(Event::<T>::BeneficiaryRemoved(staker, contract_id));
+            Ok(().into())
+        }
+
+        /// Updates the beneficiary to a new account.
+        ///
+        /// The dispatch origin is the current beneficiary.
+        #[pallet::weight(10_000)]
+        pub fn update_rewards_beneficiary(
```
Dinonard, line 1035: Perhaps `delegate_rewards_beneficiary`?
Since this cannot be used by the staker, so it's more clear.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -795,18 +815,24 @@ pub mod pallet {
...
-            Self::deposit_event(Event::<T>::Reward(staker, contract_id, era, staker_reward));
+            Self::deposit_event(Event::<T>::Reward(reward_account, contract_id, era, staker_reward));
 
-            Ok(Some(if should_restake_reward {
-                T::WeightInfo::claim_staker_with_restake()
-            } else {
-                T::WeightInfo::claim_staker_without_restake()
-            })
-            .into())
+            Ok(Some(weight_info).into())
```
Dinonard, line 835: Is this a personal preference or was the approach before incorrect?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -795,18 +815,24 @@ pub mod pallet {
                     contract_id.clone(),
                     staker_reward,
                 ));
+
+                weight_info = T::WeightInfo::claim_staker_with_restake();
             }
 
-            T::Currency::resolve_creating(&staker, reward_imbalance);
+            // Withdraw reward funds from the dapps staking pot
+            let reward_imbalance = T::Currency::withdraw(
```
Dinonard, line 823: What if this fails at this point?
What state do we leave the chain in?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -751,7 +778,10 @@ pub mod pallet {
                 staker_info.latest_staked_value(),
             );
 
-            if should_restake_reward {
+            let beneficiary = RewardsBeneficiary::<T>::get(&staker, &contract_id);
+            let mut weight_info = T::WeightInfo::claim_staker_without_restake();
+
+            if beneficiary.is_none() && should_restake_reward {
```
Dinonard, line 784: I'm not fond of argument bloat, but perhaps the added check would be better suited inside `fn should_restake_reward`?

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -795,18 +815,24 @@ pub mod pallet {
...
+                staker_reward,
+                WithdrawReasons::TRANSFER,
+                ExistenceRequirement::AllowDeath,
+            )?;
+
+            let reward_account = beneficiary.clone().unwrap_or(staker.clone());
+            T::Currency::resolve_creating(&reward_account, reward_imbalance);
             Self::update_staker_info(&staker, &contract_id, staker_info);
-            Self::deposit_event(Event::<T>::Reward(staker, contract_id, era, staker_reward));
+            Self::deposit_event(Event::<T>::Reward(reward_account, contract_id, era, staker_reward));
```
Dinonard, line 833: Perhaps this is confusing a bit - `Reward` event would be read as `X staked on Y and received Z amount of reward for era N`.

Right now, beneficiary could be an account that's not even part of dapps-staking, so it'd be even more confusing.

Could you share your opinion on this?
Is there an alternative solution you could propose?

frame/dapps-staking/src/tests.rs
```diff
@@ -1782,6 +1782,58 @@ fn claim_only_payout_is_ok() {
     })
 }
 
+#[test]
```
Dinonard, line 1785: Nice UT coverage! :+1:

Dinonard (commented):
Great work, that was super fast :smiley: 

I've left some comments/questions and would like to hear your opinion.

frame/dapps-staking/src/pallet/mod.rs
kaichaosun: It's a personal preference, trying to make return logic simple and default based.

frame/dapps-staking/src/pallet/mod.rs
kaichaosun: The failed call will not affect the chain state, since the call by default is transactional.

frame/dapps-staking/src/pallet/mod.rs
kaichaosun: Sounds good!

frame/dapps-staking/src/pallet/mod.rs
kaichaosun: Current Reward event reads as `someone staked on Y and its beneficiary is X, X received Z amount of reward for era N.` 
Adding an extract field called `staker` makes sense for easily index on staker for reward events.

For UX, most users don't need to bother this beneficiary settings, by default the `restake` should just works fine. User only need to care about beneficiary if they are making a staking pool for others, delegate and beneficiary will make it clear in this case.

Instead of adding an extra storage, make the reward destination to include beneficiary seems more straight forward, but require more thorough design and changes.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: You're right, after https://github.com/paritytech/substrate/pull/11431.

frame/dapps-staking/src/pallet/mod.rs
Dinonard: Sure, I like the scenarios, even though this is just a dummy-pre-screening-task feature :)

Differentiating when stakers receive rewards and when it's just via a reward beneficiary is important if we want to be aware how much someone earned as a staker. With the modified event, we cannot tell whether the user received rewards as a beneficiary or as a nomination reward.

And love the proposal about reward destination :+1: :smile: , was hoping for it!

Dinonard: Sure, I like the scenarios, even though this is just a dummy-pre-screening-task feature :)

Differentiating when stakers receive rewards and when it's just via a reward beneficiary is important if we want to be aware how much someone earned as a staker. With the modified event, we cannot tell whether the user received rewards as a beneficiary or as a nomination reward.

And love the proposal about reward destination :+1: :smile: , was hoping for it!

PR #114, opened 2022-11-15
Custom barrier for DescendOrigin

Dinonard:
**Pull Request Summary**

Adds custom `Barrier` for `XCM` that allows sequence like:

`DescendOrigin -> WithdrawAsset -> BuyExecution`

This is required if we want to support remote execution from non-parachain accounts.
The common sense here is the assumption that none of the parachains would allow sending of customizable XCM sequence
that isn't prepended with `DescendOrigin`.

**Check list**
- [x] added unit tests
- [x] updated documentation

### DIFF
--- a/primitives/xcm/src/lib.rs
+++ b/primitives/xcm/src/lib.rs
@@ -256,10 +256,8 @@ impl<
 /// payments into account.
 ///
 /// Only allows for sequence `DescendOrigin` -> `WithdrawAsset` -> `BuyExecution`
-pub struct AllowTopLevelPaidExecutionWithDescendOriginFrom<T>(PhantomData<T>);
-impl<T: Contains<MultiLocation>> ShouldExecute
-    for AllowTopLevelPaidExecutionWithDescendOriginFrom<T>
-{
+pub struct AllowPaidExecWithDescendOriginFrom<T>(PhantomData<T>);
+impl<T: Contains<MultiLocation>> ShouldExecute for AllowPaidExecWithDescendOriginFrom<T> {
     fn should_execute<RuntimeCall>(
         origin: &MultiLocation,
         message: &mut Xcm<RuntimeCall>,
@@ -268,7 +266,7 @@ impl<T: Contains<MultiLocation>> ShouldExecute
     ) -> Result<(), ()> {
         log::trace!(
             target: "xcm::barriers",
-            "AllowTopLevelPaidExecutionWithDescendOriginFrom origin: {:?}, message: {:?}, max_weight: {:?}, weight_credit: {:?}",
+            "AllowPaidExecWithDescendOriginFrom origin: {:?}, message: {:?}, max_weight: {:?}, weight_credit: {:?}",
             origin, message, max_weight, _weight_credit,
         );
         ensure!(T::contains(origin), ());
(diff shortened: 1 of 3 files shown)


primitives/xcm/src/lib.rs
```diff
@@ -250,3 +251,56 @@ impl<
...
+        log::trace!(
+            target: "xcm::barriers",
+            "AllowPaidExecWithDescendOriginFrom origin: {:?}, message: {:?}, max_weight: {:?}, weight_credit: {:?}",
+            origin, message, max_weight, _weight_credit,
+        );
+        ensure!(T::contains(origin), ());
+        let mut iter = message.0.iter_mut();
+        let mut i = iter.next().ok_or(())?;
+
+        match i {
```
0x7CFE: If I'm not mistaken, this whole sequence can be replaced with a single `assert_matches!`. Or, if it's not available in this code, a match on an array.

Something like this:
```rust
match iter.take(3).collect::<Vec<_>>() {
    [ DescendOrigin(..), WithdrawAsset(..), BuyExecution { weight_limit, .. } ] => { /* check weight_limit */ }
    _ => return Err(())
}
```

Dinonard: I haven't seen `assert_matches!` used outside of unit test context so would like to avoid it in runtime code.

For the suggestion you made, I did it like this:
```Rust
        let iter = message.0.iter_mut();

        match iter.take(3).collect::<Vec<_>>()[..] {
            [DescendOrigin(..), WithdrawAsset(..), BuyExecution {
                ref mut weight_limit,
                ..
            }] => {
                match weight_limit {
                    Limited(ref mut weight) if *weight >= max_weight => {
                        *weight = max_weight;
                        Ok(())
                    }
                    Unlimited => {
                        *weight_limit = Limited(max_weight);
                        Ok(())
                    }
                    _ => Err(()),
                }
            }
            _ => return Err(()),
        }
```
I feel as if few extra lines were saved but readability has decreased significantly.

0x7CFE: Agree. This version do look awkward. I missed that we need to assign `weight_limit` back.

0x7CFE: Out of curiosity, would something like this work? Don't sure if it would compile, though.

```rust
        match iter.take(3).collect::<Vec<_>>()[..] {
            [DescendOrigin(..), WithdrawAsset(..), BuyExecution { weight_limit: Limited(ref mut limit), .. }] if *limit >= max_weight => {
                *limit = max_weight;
                Ok(())
            }

            [DescendOrigin(..), WithdrawAsset(..), BuyExecution { weight_limit: ref mut limit @ Unlimited, .. }] => {
                *limit = Limited(max_weight);
                Ok(())
            }

            _ => return Err(()),
        }    
```

P.S.: You can use `.as_slice()` instead of `[..]`

Dinonard: Comments implemented.

Also, paranoia kicked in so I added additional UTs.

Dinonard: @0x7CFE if no more comments, can you please approve?

0x7CFE (approved):
One question though: why do we consider a sequence as valid if it contains some extra instructions? I mean, technically we can, of course, but should we?

Dinonard:
> One question though: why do we consider a sequence as valid if it contains some extra instructions? I mean, technically we can, of course, but should we?

Because the point of this barrier is to ensure that whoever is sending the sequence pays for it's execution.
Whatever they do after paying is fine - that's why the UT contains scenario with one extra instruction.

The sequence being matched in the code is useless by itself.

PR #115, opened 2022-11-21
Prevent double & incorrect charging of EVM-XVM gas fee

Dinonard:
**Pull Request Summary**

By specifying `max_fee_per_gas` to be `zero`, we instruct EVM executor not to charge the caller for gas fees.

This is ok since the XVM call will be charged in the context of its originating call.

EVM runner reference code: https://github.com/paritytech/frontier/blob/master/frame/evm/src/runner/stack.rs#L135

### DIFF
--- a/frame/pallet-xvm/src/evm.rs
+++ b/frame/pallet-xvm/src/evm.rs
@@ -27,6 +27,8 @@ where
             from, to, input,
         );
         let value = U256::zero();
+
+        // Tells the EVM executor that no fees should be charged for this execution.
         let max_fee_per_gas = U256::zero();
         let gas_limit = T::GasWeightMapping::weight_to_gas(context.max_weight);
         log::trace!(
--- a/frame/pallet-xvm/src/evm.rs
+++ b/frame/pallet-xvm/src/evm.rs
@@ -1,9 +1,9 @@
 //! EVM support for XVM pallet.
 
 use crate::*;
-use pallet_evm::GasWeightMapping;
+use pallet_evm::{GasWeightMapping, Runner};
 use sp_core::{H160, U256};
-use sp_runtime::traits::Get;
+use sp_runtime::traits::{Get, UniqueSaturatedInto};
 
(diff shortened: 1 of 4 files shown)


0x7CFE (approved):

PierreOssun (approved):
LGTM

frame/pallet-xvm/src/evm.rs
```diff
@@ -26,8 +26,10 @@ where
             "Start EVM XVM: {:?}, {:?}, {:?}",
             from, to, input,
         );
-        let value = U256::from(0u64);
-        let max_fee_per_gas = U256::from(3450898690u64);
+        let value = U256::zero();
+
+        // Tells the EVM executor that no fees should be charged for this execution.
+        let max_fee_per_gas = U256::zero();
```
akru, line 32: Is it really works? I guess EVM has minimal fee and return an error if providen fee is less.

Dinonard: I believe it will work.

Calculated initial fee will be `zero`.

```
	fn withdraw_fee(who: &H160, fee: U256) -> Result<Self::LiquidityInfo, Error<T>> {
		if fee.is_zero() {
			return Ok(None);   <===========
		}
               ...
```

Followed by:
```
fn correct_and_deposit_fee(
		who: &H160,
		corrected_fee: U256,
		base_fee: U256,
		already_withdrawn: Self::LiquidityInfo,
	) -> Self::LiquidityInfo {
		if let Some(paid) = already_withdrawn {
                    .....
		}
		None   <========
	}
```

So the whole payment will be a no-op.

Dinonard: @akru it seems that is correct - this step happens in transaction validation.

We'll have to rethink how we approach this - probably calling `EVM` pallet isn't the way to go.
However, the approach with `let max_fee_per_gas = U256::zero();` does mean that no fees will actually be consumed from the caller.

BUT we need to bypass the transaction validation...

PR #118, opened 2022-11-28
Custom sig filter bypass

Dinonard:
**Pull Request Summary**

Fix the issue where call filter would by bypassed with custom sig.

**Check list**
- [x] added unit tests


akru (approved):
LGTM

PR #121, opened 2022-12-19
Burn stale rewards

Dinonard:
**Pull Request Summary**

This is in response to request made by **ShidenDAO** to support burning of _stale_ developer rewards.

In case developer's contract gets unregistered, they can still claim their old rewards.
Before this update, this was possible indefinitely, and some devs would never do this.

With this update, a new extrinsic call is introduced which will burn unregistered contract's unclaimed rewards.

**Summary:**
* the call needs to be made with `root` privileges
* in case contract is still registered, the call will fail
* in case reward was already claimed, it cannot be burned
* once burned, reward cannot be claimed
* the burn is executed per era (same as rewards are claimed)
  * in case multiple eras need burning, multiple calls will be required (e.g. via batch call)

**Check list**
- [x] added unit tests
- [x] updated documentation

### DIFF
--- a/frame/dapps-staking/src/lib.rs
+++ b/frame/dapps-staking/src/lib.rs
@@ -34,6 +34,7 @@
 //! - `maintenance_mode` - enables or disables pallet maintenance mode
 //! - `set_reward_destination` - sets reward destination for the staker rewards
 //! - `set_contract_stake_info` - root-only call to set storage value (used for fixing corrupted data)
+//! - `burn_stale_reward` - root-only call to burn unclaimed, stale rewards from unregistered contracts
 //!
 //! User is encouraged to refer to specific function implementations for more comprehensive documentation.
 //!
@@ -122,6 +123,15 @@ impl<AccountId> DAppInfo<AccountId> {
         Self {
             developer,
             state: DAppState::Registered,
+        }
+    }
+
+    /// `true` if dApp has been unregistered, `false` otherwise
+    fn is_unregistered(&self) -> bool {
+        if let DAppState::Unregistered(_) = self.state {
+            true
+        } else {
+            false
         }
     }
 }
--- a/frame/dapps-staking/src/lib.rs
+++ b/frame/dapps-staking/src/lib.rs
@@ -128,11 +128,7 @@ impl<AccountId> DAppInfo<AccountId> {
 
     /// `true` if dApp has been unregistered, `false` otherwise
     fn is_unregistered(&self) -> bool {
-        if let DAppState::Unregistered(_) = self.state {
-            true
-        } else {
-            false
-        }
+        matches!(self.state, DAppState::Unregistered(_))
     }
 }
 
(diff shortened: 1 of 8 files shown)


frame/dapps-staking/src/lib.rs
```diff
@@ -124,6 +125,15 @@ impl<AccountId> DAppInfo<AccountId> {
...
+
+    /// `true` if dApp has been unregistered, `false` otherwise
+    fn is_unregistered(&self) -> bool {
+        if let DAppState::Unregistered(_) = self.state {
+            true
+        } else {
+            false
+        }
+    }
 }
```
0x7CFE:
```suggestion
    fn is_unregistered(&self) -> bool {
        std::matches!(self.state, DAppState::Unregistered(_))
    }
}
```

0x7CFE (approved):
LGTM

frame/dapps-staking/src/lib.rs
akru: I believe `std` may be unavailable for `no_std`.

frame/dapps-staking/src/pallet/mod.rs
```diff
@@ -918,9 +911,92 @@ pub mod pallet {
...
+            let dapp_reward = Self::calculate_dapp_reward(&contract_stake_info, &dapp_info, era)?;
+
+            // Withdraw reward funds from the dapps staking pot and burn them
+            let imbalance_to_burn = T::Currency::withdraw(
+                &Self::account_id(),
+                dapp_reward,
+                WithdrawReasons::TRANSFER,
+                ExistenceRequirement::AllowDeath,
+            )?;
+            mem::drop(imbalance_to_burn);
```
akru, line 950: De we really need to do it? Local variable automatically will be removed after out of scope.

Dinonard: No, it's not needed explicitly.
I tested before without it and it worked just fine.

But I wanted to be explicit so it's more clear what I'm doing.
I also found the `burn_from` functionality but had some issues with it. Will look into it again.

EDIT: The code will end up looking more complicated so I will avoid it. The `burn_from` also deposits `Withdraw` event so what I'm doing now is basically the same.

frame/dapps-staking/src/lib.rs
Dinonard: `matches!` is available in `sp-std` so I will just use it from there.

Good suggestion!

frame/dapps-staking/src/pallet/mod.rs
akru: `let _ = T::Currency::withdraw(`

Should works the same.

Dinonard: Yes, I implemented it initially almost like that.
It was `let _reward_to_burn = ...` :slightly_smiling_face: 

I added the `drop` later since I wanted to be more explicit.
We can change it in future PR, I don't really mind.

PR #127, opened 2023-02-02
Assets-pallet chain-extension

PierreOssun:
**Pull Request Summary**

This PR adds a 1 on 1 map of pallet-assets functions to chain-extension.
For each extrinsic:
- charge the weight before entering the pallet call
- the origin signer is defined by an `Origin` enum that takes either `Caller` or `Contract` as value. This way the contract can chose between making call on behalf of the caller or on behalf of itself
- return a Retval (0 for success or an error number that will be decoded on contract side)

integrations tests here:
https://github.com/AstarNetwork/chain-extension-contracts/pull/11

**Check list**
- [x] implemented chain-extension


chain-extensions/types/assets/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+
+impl TryFrom<DispatchError> for AssetsError {
+    type Error = DispatchError;
+
+    fn try_from(input: DispatchError) -> Result<Self, Self::Error> {
+        let error_text = match input {
+            DispatchError::Module(ModuleError { message, .. }) => message,
+            _ => Some("No module error Info"),
+        };
+        return match error_text {
```
Dinonard, line 85: Not related to this PR in particular, but it might be good to have some integration test to ensure change in pallet doesn't break the CE. Right now only a string, with no bounds to `pallet-assets` `Error` enum exists.

Or perhaps this could be changed to something like `pallet_assets::Error::<T>::BalanceLow.to_string()`?
That way we can rely on compiler to detect if some enum value is renamed.

chain-extensions/types/assets/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Astar. If not, see <http://www.gnu.org/licenses/>.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use scale_info::scale;
+use sp_runtime::{DispatchError, ModuleError};
+
+#[derive(PartialEq, Eq, Copy, Clone, scale::Encode, scale::Decode, Debug)]
```
Dinonard: Just an opinion - might look nice to import types from `scale` via `use ...` instead.

chain-extensions/types/assets/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+pub enum AssetsError {
+    /// Success
+    Success = 0,
```
Dinonard: Strange to call the enum `Error` :slightly_smiling_face: 

Maybe something like `Outcome`? Just a suggestion.

chain-extensions/pallet-assets/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+                    T::Balance,
+                ) = env.read_as()?;
+
+                let base_weight = <T as pallet_assets::Config>::WeightInfo::create();
+                env.charge_weight(base_weight)?;
+
+                let runtime_origin = RawOrigin::Signed(match origin {
+                    Origin::Caller => env.ext().caller().clone(),
+                    Origin::Address => env.ext().address().clone(),
+                });
```
Dinonard: Just to check - this isn't controlled directly by the caller but instead by context, right?

E.g. I cannot call this function as someone else :slightly_smiling_face:

Dinonard (commented):
LGTM

Just a few small comments/questions.

chain-extensions/pallet-assets/src/lib.rs
PierreOssun: It's in the context. Chain-extension can only be called from contract. And in the contract call, the contract chose the origin 
https://github.com/AstarNetwork/chain-extension-contracts/blob/eb1b6892ef1ab002a8694f1f59678f0c138f2de3/examples/assets-mock/lib.rs#L39

chain-extensions/pallet-assets/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+                    Origin,
+                    <T as pallet_assets::Config>::AssetId,
+                    T::AccountId,
+                    T::Balance,
+                ) = env.read_as()?;
+
+                let base_weight = <T as pallet_assets::Config>::WeightInfo::transfer();
+                env.charge_weight(base_weight)?;
+
+                let runtime_origin = RawOrigin::Signed(match origin {
```
akru: May be having some kind of template or function for this match could be useful for improving readability.

chain-extensions/pallet-assets/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+                        let mapped_error = Outcome::from(e);
+                        Ok(RetVal::Converging(mapped_error as u32))
+                    }
+                    Ok(_) => Ok(RetVal::Converging(Outcome::Success as u32)),
+                };
+            }
+            AssetsFunc::MetadataName => {
+                let id: <T as pallet_assets::Config>::AssetId = env.read_as()?;
+
+                let base_weight = <T as frame_system::Config>::DbWeight::get().reads(1);
```
akru: Better to have it in same manner with other weight getters, magic numbers is not best practice.

akru (changes requested):
My proposal is rebase it, this PR brings a lot of not related commits.

Also provided some style improvements.

chain-extensions/pallet-assets/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+            14 => Ok(AssetsFunc::MetadataDecimals),
+            _ => Err(DispatchError::Other(
+                "PalletAssetsExtension: Unimplemented func_id",
+            )),
+        }
+    }
+}
+
+/// Pallet Assets chain extension.
+pub struct AssetsExtension<T>(PhantomData<T>);
```
akru: I believe weight info should be generic type parameter that implements your defined ChainExtensionWeightInfo trait.

because chain extensions have no Config trait implementation (as it done for pallets). To push custom weight implementations we should use type magic for CE structure.

in your case it will be `AssetExtension<T, W>`, and as usual we can define `()` implementation for this trait for making it easy to use by default.

akru (approved):

PierreOssun:
Closed this one
Reopen on #132

PR #130, opened 2023-02-07
The account abstraction pallet

akru:
**Pull Request Summary**

Account abstraction pallet (`pallet-account`) makes possible to manage multiple origins and proxy call using desired account origin.

**Check list**
- [x] added unit tests
- [ ] updated documentation


shunsukew:
~~Can I ask what are purposes and use cases of `call_as` call to dispatch by using derived account? I'm interested in it.~~

frame/pallet-account/src/pallet/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+        pub fn meta_call(
+            origin: OriginFor<T>,
+            call: Box<<T as Config>::RuntimeCall>,
+            signer: T::AccountId,
+            signature: T::Signature,
+        ) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+
+            let magic = T::ChainMagic::get();
+            let nonce = frame_system::Pallet::<T>::account_nonce(signer.clone());
```
shunsukew: Signer usually doesn't have enough balance, and sometimes doesn't have it at all.
How do we hold nonce of accounts without Existential Deposit (ED)?

And because of ED, account info can be cleaned up, which means account nonce can be reset to 0, so I think it cannot ensure replay attack when that account is recreated later.

Being discussed here https://github.com/paritytech/substrate/issues/11988

frame/pallet-account/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Astar. If not, see <http://www.gnu.org/licenses/>.
+
+//! # Account abstraction pallet
+//!
+//! ## Overview
+//!
+//! An accout abstraction pallet make possible to derive new blockchain based
```
0x7CFE:
```suggestion
//! An accout abstraction pallet makes it possible to derive new blockchain based
```

frame/pallet-account/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+//! account for your existed external owned account (seed phrase based). The onchain
```
0x7CFE:
```suggestion
//! account for an existing external owned account (seed phrase based). The onchain
```

frame/pallet-account/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+//! account could be drived to multiple address spaces: H160 and SS58. For example,
```
0x7CFE:
```suggestion
//! account could be derived to multiple address spaces: H160 and SS58. For example,
```

frame/pallet-account/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+//! it makes possible predictable interaction between substrate native account and
+//! EVM smart contracts.
```
0x7CFE:
```suggestion
//! it makes possible to predictably interact between substrate native account and
//! EVM smart contracts.
```

frame/pallet-account/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+pub mod pallet;
+pub use pallet::pallet::*;
+
+pub mod weights;
+pub use weights::*;
+
+#[cfg(test)]
+mod mock;
+#[cfg(test)]
+mod tests;
```
Dinonard, line 52: Why not just add the logic from the _pallet/mod.rs_ here? Seems redundant to have that module/file.

frame/pallet-account/src/origins.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// Derive new origin.
+pub trait OriginDeriving<AccountId, Origin> {
+    /// Derive new origin depend of account and index
+    fn derive(&self, source: &AccountId, index: u32) -> Origin;
+}
+
+/// Origin that support native and EVM compatible options.
+#[derive(PartialEq, Eq, Clone, RuntimeDebug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+pub enum NativeAndEVM {
```
Dinonard, line 31: Perhaps a more general name which is less _restrictive_ to potential future types?

frame/pallet-account/src/origins.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+}
+
+/// Kind for NativeAndEVM origin.
+#[derive(PartialEq, Eq, Clone, RuntimeDebug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+pub enum NativeAndEVMKind {
+    Native,
+    H160,
+}
```
Dinonard, line 63: Perhaps adding `impl Into` for `enum NativeAndEVM` that converts into `NativeAndEVMKind` is a good idea - you can test the conversion & ensure it's exhaustive.

frame/pallet-account/src/pallet/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+    use frame_system::pallet_prelude::*;
+    use sp_runtime::traits::{IdentifyAccount, Verify};
+
+    /// The current storage version.                                                                                      
+    const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
+    #[pallet::pallet]
+    #[pallet::generate_store(pub(super) trait Store)]
+    #[pallet::storage_version(STORAGE_VERSION)]
+    #[pallet::without_storage_info]
```
Dinonard: This should be removed and pallet should correctly be _bounded_ in relations to PoV size.

frame/pallet-account/src/pallet/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+        },
+    }
+
+    #[pallet::origin]
+    pub type Origin<T> = <T as Config>::CustomOrigin;
+
+    /// Account origins
+    #[pallet::storage]
+    pub type AccountOrigin<T: Config> =
+        StorageMap<_, Blake2_128Concat, T::AccountId, Vec<T::CustomOrigin>, ValueQuery>;
```
Dinonard: `Vec<_>` should not be used in storage. Instead, you should use `BoundedVec`. That way you can also get rid of the `without_storage_info` macro you used above. This item cannot be _unlimited_.

frame/pallet-account/src/pallet/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+            origin_kind: T::CustomOriginKind,
+        ) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+
+            let mut origins = AccountOrigin::<T>::get(who.clone());
+            let next_index = origins.len();
+            let new_origin = origin_kind.derive(&who, next_index as u32);
+
+            origins.push(new_origin.clone());
+            AccountOrigin::<T>::insert(who.clone(), origins);
```
Dinonard: There should be a price for this - I'd say storage rent but given it's not a _rent_ since derived accounts cannot be deleted, it's more appropriate to just charge users some amount (and deposit it into treasury or burn it).

frame/pallet-account/src/pallet/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+        ///
+        /// The dispatch origin for this call must be _Signed_.
+        ///
+        /// Parameters:
+        /// - `origin_index`: Account origin index for using.
+        /// - `call`: The call to be made by the `derived` account.
+        #[pallet::weight({
+			let di = call.get_dispatch_info();
+			(T::WeightInfo::proxy_call()
+				.saturating_add(T::DbWeight::get().reads_writes(1, 1))
```
Dinonard: Why is the adding of 1 R/W weight needed?
`T::WeightInfo::proxy_call` should capture the required weight (for e.g. `System::remark` call).

frame/pallet-account/src/pallet/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(crate) fn deposit_event)]
+    pub enum Event<T: Config> {
+        NewOrigin {
+            account: T::AccountId,
+            origin: T::CustomOrigin,
+        },
+        ProxyCall {
+            payer: T::AccountId,
```
Dinonard: Why is `payer` needed here?

frame/pallet-account/src/pallet/mod.rs
Dinonard: Agree with Shunsuke, `meta_call` should be left for some other PR in the future, potentially.

frame/pallet-account/src/pallet/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+            let next_index = origins.len();
+            let new_origin = origin_kind.derive(&who, next_index as u32);
+
+            origins.push(new_origin.clone());
+            AccountOrigin::<T>::insert(who.clone(), origins);
+
+            Self::deposit_event(Event::NewOrigin {
+                account: who,
+                origin: new_origin,
+            });
```
Dinonard: Just wondering - why do we even need to store the derived address?

If user would specify Origin kind & index, we could always derive the same address. Is there some security risk that comes with such approach?

Not sure whether it's more computationally expensive to derive the address or read the DB entry.

frame/pallet-account/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "pallet-account"
+authors = ["Stake Technologies"]
+edition = "2021"
```
Dinonard: Please use workspace inheritance for relevant package info values.

frame/pallet-account/src/weights.rs
```diff
@@ -0,0 +1,37 @@
...
+
+// Astar is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Astar. If not, see <http://www.gnu.org/licenses/>.
+
+use frame_support::weights::Weight;
```
Dinonard, line 19: I don't see any TODO for this, but you should also add benchmarking code.

Dinonard (changes requested):

frame/pallet-account/src/origins.rs
shunsukew: `CustomOrigin` and `CustomOriginKind` are the ones generic.
I guess this is because of tests concrete enum type. but I agree with something more general

frame/pallet-account/src/pallet/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+        /// Weight information for extrinsics in this pallet.
+        type WeightInfo: WeightInfo;
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        /// Origin with given index not registered.
+        UnregisteredOrigin,
+        /// Signer not match signature, check nonce, magic and try again.
```
PierreOssun:
```suggestion
        /// Signature does not match Signer, check nonce, magic and try again.
```

frame/pallet-account/src/pallet/mod.rs
akru: In case of store derived address we can charge origin creation fee, it looks more secure because it prevents address derivation manipulation using GPU, etc.

frame/pallet-account/src/origins.rs
akru: This is just example, we can add custom enum for each runtime.

frame/pallet-account/src/lib.rs
akru: Moved

frame/pallet-account/src/pallet/mod.rs
akru: Temporary removed

frame/pallet-account/src/pallet/mod.rs
akru: Rewritten to make it bounded.

frame/pallet-account/src/pallet/mod.rs
akru: Used double-map instead of vector.

frame/pallet-account/src/pallet/mod.rs
akru: Removed

frame/pallet-account/Cargo.toml
akru: Added

frame/pallet-account/src/origins.rs
akru: Probably yes, but may be we can generate *Kind structure from custom origin somehow. I'll investigate later.

akru:
@Dinonard please check changes, I'll happy to merge it soon

frame/pallet-account/src/tests.rs
```diff
@@ -0,0 +1,81 @@
...
+            System::events()
+                .last()
+                .expect("events expected")
+                .event
+                .clone(),
+            RuntimeEvent::Account(Event::ProxyCall{origin, ..})
+            if origin == NativeAndEVM::Native(ALICE_D1_NATIVE.into())
+        );
+    })
+}
```
Dinonard, line 145: You should also add some negative tests to cover the possible errors.

frame/pallet-account/src/weights.rs
Dinonard: @akru this comment is unresolved.

frame/pallet-account/src/pallet/mod.rs
Dinonard: @akru This comment is still relevant.

frame/pallet-account/src/pallet/mod.rs
Dinonard: @akru This comment is still relevant.

frame/pallet-account/src/weights.rs
akru: Added TODO, I'll add weights in next PR.

frame/pallet-account/src/tests.rs
akru: Added

frame/pallet-account/src/weights.rs
akru: Added benchmarks, next step generate weights

frame/pallet-account/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+    #[pallet::pallet]
+    #[pallet::generate_store(pub(super) trait Store)]
+    #[pallet::storage_version(STORAGE_VERSION)]
+    pub struct Pallet<T>(PhantomData<T>);
+
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+        /// Custom origin type.
+        type CustomOrigin: Parameter + TryInto<Self::AccountId> + MaxEncodedLen;
+        /// Parameter that defin different origin options and how to create it.
```
shaunxw:
```suggestion
        /// Parameter that defines different origin options and how to create it.
```

frame/pallet-account/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+    impl<T: Config> Pallet<T> {
+        /// Derive new origin for account.
+        ///
+        /// The dispatch origin for this call must be _Signed_.
+        #[pallet::weight(
+            T::WeightInfo::new_origin()
+                .saturating_add(T::DbWeight::get().reads_writes(1, 2))
+        )]
+        #[pallet::call_index(0)]
+        pub fn new_origin(
```
shaunxw, line 154: As the origin derivation algorithm is deterministic, could we skip the origin registering? The derived origin could be derived just in time in a `proxy_call`, like:

```rust
Account::proxy_call(alice, derive_kind, index, call);
```

If we do need to cache the derived origin to reduce gas costs, the cache should be cleaned on kill account.

frame/pallet-account/src/origins.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl OriginDeriving<AccountId32, NativeAndEVM> for NativeAndEVMKind {
+    fn derive(&self, source: &AccountId32, index: u32) -> NativeAndEVM {
+        let salted_source = [source.as_ref(), &index.encode()[..]].concat();
+        let derived = sp_io::hashing::blake2_256(&salted_source);
+        match self {
+            NativeAndEVMKind::Native => NativeAndEVM::Native(derived.into()),
+            NativeAndEVMKind::H160 => NativeAndEVM::H160(sp_core::H160::from_slice(&derived[..20])),
+        }
+    }
```
shaunxw, line 73: Dumb question: is our custom origin deriving meant to derive between H160 and Native? The impl here seems only to derive sub-accounts from their own type.

frame/pallet-account/src/origins.rs
```diff
@@ -0,0 +1,74 @@
...
+/// Kind for NativeAndEVM origin.
+#[derive(PartialEq, Eq, Clone, RuntimeDebug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+pub enum NativeAndEVMKind {
+    Native,
+    H160,
+}
+
+impl OriginDeriving<AccountId32, NativeAndEVM> for NativeAndEVMKind {
+    fn derive(&self, source: &AccountId32, index: u32) -> NativeAndEVM {
+        let salted_source = [source.as_ref(), &index.encode()[..]].concat();
```
shaunxw, line 67: ~I think we can add module id as entropy, probably same to module id in pallet-utilities, to be compatible. https://github.com/paritytech/substrate/blob/f4a2e84ee5974b219f2a03cd195105060c41e3cd/frame/utility/src/lib.rs#LL506C28-L506C28~

EDIT: probably not a good idea to be compatible with pallet-utilities.  Compatibility with main stream derivation standard is more important

frame/pallet-account/src/tests.rs
```diff
@@ -0,0 +1,114 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Astar. If not, see <http://www.gnu.org/licenses/>.
+
+use super::*;
+use assert_matches::assert_matches;
+use frame_support::assert_ok;
+use mock::*;
+
+#[test]
+pub fn new_origin_works() {
```
shaunxw, line 25: Will be good to add event checks.

frame/pallet-account/src/tests.rs
```diff
@@ -0,0 +1,114 @@
...
+            System::events()
+                .last()
+                .expect("events expected")
+                .event
+                .clone(),
+            RuntimeEvent::Account(Event::ProxyCall{origin, ..})
+            if origin == NativeAndEVM::Native(ALICE_D1_NATIVE.into())
+        );
+    })
+}
```
shaunxw, line 112: IMO the origin of the balance transfer should also be checked, probably via transfer event.

frame/pallet-account/src/lib.rs
akru: I believe derivation algorithm is just a sample, it not restricted to be deterministic. Account registration helps to prevent spamming and account address manipulation.

frame/pallet-account/src/origins.rs
akru: Yes, to create new origin the call origin must be substrate-compatible.

frame/pallet-account/src/tests.rs
akru: Let me add some.

frame/pallet-account/src/tests.rs
akru: Sure, added, please check @shaunxw

frame/pallet-account/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+            origin: OriginFor<T>,
+            origin_kind: T::CustomOriginKind,
+        ) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+
+            let next_index = AccountLastOrigin::<T>::get(&who);
+            let new_origin = origin_kind.derive(&who, next_index);
+            AccountOrigin::<T>::insert(&who, next_index, new_origin.clone());
+            AccountLastOrigin::<T>::insert(&who, next_index + 1);
```
Dinonard, line 171: This is still free - so my comment from 2 months ago is still valid.
You should charge  _rent fee_ for these two items since you're inflating the database.

Check how `pallet-assets` is doing it, or any other pallet in `substrate` that creates some storage items (e.g. `pallet-identity`, `pallet-multisig`, etc.).

Dinonard (changes requested):

frame/pallet-account/src/lib.rs
akru: Technically, you pay transaction fee to cover it (prevent spam), it’s not required at all, just one of ways to do it. I’m not sure it’s required for the prototype.

akru:
Thank you for your feedback, @shaunxw! If you have no more concerns, then let’s approve and merge this PR.

frame/pallet-account/src/lib.rs
akru: My personal opinion - better charge fee than just lock token forever. Charged fee much more controllable, it could be burn, redistributed or whatever you want. But locked deposit is just locked.

frame/pallet-account/src/origins.rs
akru: Honestly, at first stage I prefer to give some sample of usage and no more. What’s about derivation standard, it could be easily implemented outside of the pallet.

frame/pallet-account/src/lib.rs
Dinonard: The way it's implemented now - sure, that's viable option. It still needs to be added though since it's missing now.

frame/pallet-account/src/lib.rs
Dinonard: @akru what is the usecase behind allowing single account to have multiple derived address of the same type?

The way I see it, it just adds unnecessary complexity.
If we only allow one account type to be derived from the origin, then it's much simpler.
If an account wants to control multiple different derived accounts, it can use proxy to achieve that.

Dinonard: Shaun also commented above([here](https://github.com/AstarNetwork/astar-frame/pull/130#discussion_r1197707974)) about what happens when account is _killed_. Without any locked or reserved funds, these storage items will remain in the DB even without an owner.

frame/pallet-account/src/lib.rs
akru: Not sure something is missing now, what precisely you are talking about?

Dinonard: I'm talking about payment for the created storage.

> Technically, you pay transaction fee to cover it (prevent spam), it’s not required at all, just one of ways to do it. I’m not sure it’s required for the prototype.

Given that our handling of fees is copy/paste of Polkadot, it is how we do things, isn't it?
Don't understand why this pallet should be any different.
Are you proposing we mix it up, some pallets charging for rent fee, while some don't? 🤷‍♂️ 

It's been on review for months, and better to do it right immediately then have someone else fix it later, add storage migration logic, etc. Doesn't matter if it's prototype or not, IMO.

PR #132, opened 2023-02-20
Assets-pallet chain-extension

PierreOssun:
**Pull Request Summary**

This PR is rebase from v0.9.37 and close previous #127 

----

This PR adds a 1 on 1 map of pallet-assets functions to chain-extension.
For each extrinsic:
- charge the weight before entering the pallet call
- the origin signer is defined by an `Origin` enum that takes either `Caller` or `Contract` as value. This way the contract can chose between making call on behalf of the caller or on behalf of itself
- return a Retval (0 for success or an error number that will be decoded on contract side)

integrations tests here:
https://github.com/AstarNetwork/chain-extension-contracts/pull/11

**Check list**
- [x] implemented chain-extension


chain-extensions/pallet-assets/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "pallet-chain-extension-assets"
+version = "0.1.0"
+authors = ["Stake Technologies <devops@stake.co.jp>"]
```
Dinonard: Aren't we? :slightly_smiling_face:

chain-extensions/pallet-assets/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+    fn default() -> Self {
+        AssetsExtension(PhantomData)
+    }
+}
+
+impl<T, W> ChainExtension<T> for AssetsExtension<T, W>
+where
+    T: pallet_assets::Config + pallet_contracts::Config,
+    <<T as SysConfig>::Lookup as StaticLookup>::Source: From<<T as SysConfig>::AccountId>,
+    <T as SysConfig>::AccountId: From<[u8; 32]>,
```
Dinonard, line 95: Can this part be removed?

chain-extensions/pallet-assets/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+
+                let base_weight = <T as pallet_assets::Config>::WeightInfo::create();
+                env.charge_weight(base_weight)?;
+
+                let call_result = pallet_assets::Pallet::<T>::create(
+                    Origin::get_origin(&origin, env).into(),
+                    id.into(),
+                    admin.into(),
+                    min_balance,
+                );
```
Dinonard: I believe this is missing the storage deposit limit check.

Creating a new asset will result in some native balance being reserved.
However, this won't be accounted for in the storage meter when compared to `storage_deposit_limit` :frowning_face: 

Looking at the chain-extension code, I feel like this is a bug in `pallet-contracts` since `storage_meter` isn't exposed. We also don't know how much balance was reserved as the result of successful `create` call.

Not sure how to tackle this.

chain-extensions/pallet-assets/src/weights.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    fn metadata_symbol() -> Weight {
+        T::DbWeight::get().reads(1 as u64)
+    }
+
+    fn metadata_decimals() -> Weight {
+        T::DbWeight::get().reads(1 as u64)
+    }
+}
```
Dinonard: Shouldn't `cargo fmt` pick this up?

chain-extensions/types/assets/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+substrate-std = [
+	"std",
+	"sp-runtime/std",
+	"frame-system/std",
+	"pallet-contracts/std",
+]
+ink-std = [
+	"std",
+	"ink/std",
+]
```
Dinonard: We had a discussion about this before but here's my new take on it.

1.
The `default` features should ensure code compiles without any issues. When doing the uplift to `polkadot-v0.9.37` I had to change `default` to `substrate-std` in order for normal `check` to pass.

2.
My experience from the uplift was that there were conflicting features (`cargo check --all-features` would fail).
It would be good to avoid this.
https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features

chain-extensions/pallet-assets/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+                    <T as pallet_assets::Config>::AssetId,
+                    T::AccountId,
+                    T::Balance,
+                ) = env.read_as()?;
+
+                let base_weight = <T as pallet_assets::Config>::WeightInfo::create();
+                env.charge_weight(base_weight)?;
+
+                let call_result = pallet_assets::Pallet::<T>::create(
+                    Origin::get_origin(&origin, env).into(),
```
Dinonard: One more comment/question about this part that I forgot - what is the use case when chain-extension acts as the original contract caller? This seems very insecure since it gives the contract essentially rights to do whatever they want with callers assets.

User can _approve_ the smart contract to move some assets using the `approve_transfer` call but that should be done via standard call to `pallet-assets::approve_transfer`.

chain-extensions/pallet-assets/src/lib.rs
PierreOssun: The create function is a payable method where you pass the amount needed. You can get this [amount with env](https://paritytech.github.io/ink/ink_env/fn.transferred_value.html).
[ink! message](https://github.com/swanky-dapps/chain-extension-contracts/blob/eb9b4f11f33c8b98156bbe4565a344f095b426e4/examples/assets/lib.rs#L24)
[test](https://github.com/swanky-dapps/chain-extension-contracts/blob/eb9b4f11f33c8b98156bbe4565a344f095b426e4/tests/assets.spec.ts#L51)

chain-extensions/types/assets/Cargo.toml
PierreOssun: If I put substrate dep in std it will not build on ink! side.
So only solution is to remove ink! dep from this crate and have a separate crate for ink!. Then we will lose coupling of Outcome enum (that was the reason why we put it in same crate)

chain-extensions/pallet-assets/src/lib.rs
PierreOssun: It won't compile if I remove this part.
I did remove it on the call function because they removed UncheckedFrom

chain-extensions/pallet-assets/src/weights.rs
PierreOssun: It seems no

chain-extensions/pallet-assets/Cargo.toml
Dinonard: Not sure why, but I read this as 'Parity' before :see_no_evil: 

Sorry for the confusion and incorrect comment, you were correct in the first place.

chain-extensions/pallet-assets/src/lib.rs
Dinonard: Ok, but there is no guarantee that the amount provided will be sufficient to pay for the asset creation, right?

E.g. a caller could pass 10^-18 ASTR and make use of 1000 ASTR available in contract to create an asset.

Chain extension should calculate the price required (just an idea), and `pallet-assets` should expose some API via which this calculation can done (doesn't seem right to copy/paste logic from pallet into chain-extension).

chain-extensions/types/assets/Cargo.toml
Dinonard: If it makes for cleaner code, perhaps it's fine.
I do believe we should follow good design principles and right now this seems a bit messy.

How about having enum generated via a simple macro?
That way we have controlled code duplication.

chain-extensions/pallet-assets/src/lib.rs
PierreOssun: I can check [value_transferred](https://github.com/paritytech/substrate/blob/38c3e5ee10be760115a77a115ef0ac6ba381542b/frame/contracts/src/exec.rs#L243) and query the deposit to check if it's "equal" ?

Dinonard: Sure, but with the changes to how the chain-extension will work (discussed yesterday), not allowing reuse of `caller` as origin to `pallet-assets` calls, then this problem is no more?

PierreOssun: It is still because the transferred balance come from caller but is balance of contract and only contract can pay for those fees

PierreOssun: Coding this makes me realize there is no need for this. As only caller will be origin of the extrinsic the contract will reserve the deposit.
If contract does not have enough balance reserve will fail and Err will be returned by the call.
If contracts does have the deposit it will be reserved.

Dinonard: Yes, that makes sense. Don't see a use case when a caller would create an asset via smart contract.

chain-extensions/pallet-assets/src/weights.rs
```diff
@@ -59,4 +60,8 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
     fn metadata_decimals() -> Weight {
         T::DbWeight::get().reads(1 as u64)
     }
+
+    fn remove_items_limit() -> Weight {
+        T::DbWeight::get().reads(1 as u64)
+    }
```
Dinonard: There's no weight associated with this, it's just a static constant.

chain-extensions/pallet-assets/src/lib.rs
```diff
@@ -364,6 +432,144 @@ where
                 let decimals = pallet_assets::Pallet::<T>::decimals(&id);
                 env.write(&decimals.encode(), false, None)?;
             }
+            AssetsFunc::StartDestroy => {
```
Dinonard: My recommendation is to remove all `destroy` related functions from the interface, because of the reasons already discussed before. This might change again in the future, leaving us with potentially incompatible chain-extension.

I'd suggest to document that in case of needing to destroy the asset, `transfer_ownership` can be used, after which required `destroy` functions can be called by an externally owned account.

chain-extensions/pallet-assets/src/lib.rs
```diff
@@ -111,8 +123,15 @@ where
                 let base_weight = <T as pallet_assets::Config>::WeightInfo::create();
                 env.charge_weight(base_weight)?;
 
+                let raw_origin = match origin {
+                    Origin::Caller => {
+                        return Ok(RetVal::Converging(Outcome::OriginCannotBeCaller as u32))
+                    }
+                    Origin::Address => RawOrigin::Signed(env.ext().address().clone()),
+                };
```
Dinonard: Please write a macro for this - it's repeated many times and given how sensitive this is, better to have it implemented exactly _once_.

chain-extensions/pallet-assets/src/weights.rs
Dinonard: Github complains newline is missing :slightly_smiling_face:

Dinonard (approved):
LGTM!

PR #139, opened 2023-03-21
`xvm_call` now returns output- or encoded error data

0x7CFE:
This PR adds the ability to get output data of a XVM call, or an error.

### Implementation details
- To get output data from pallet contracts, we now use `bare_call` which returns `ContractExecResult`
- In order to access output data from EVM _precompile_, a new `xvm_bare_call` API was added to the XVM _pallet_. 
- This allows doing direct cross pallet calls completely bypassing the `Dispatchable` API and without emitting events.
- `xvm_call` precompile now returns `(bool success, uint64 gas_spent, bytes memory data)`

### Notes
According to the design, XVM does not pose any restrictions on the format of the data returned from the call. It is considered to be an implementation detail of the XVM backend. For  example, error data returned from Wasm backend are in fact SCALE encoded `Result`. For EVM backend this can be different.

### How it was tested
- `flipper` contract from ink! examples was deployed on Shibuya/Zombienet
- Solidity contract was written that calls into flipper and gets its value
- Logs were checked to see that returned value does indeed contain value of flipper's bit

### DIFF
--- a/frame/pallet-xvm/src/lib.rs
+++ b/frame/pallet-xvm/src/lib.rs
@@ -80,6 +80,12 @@ pub struct XvmCallOk {
     consumed_weight: u64,
 }
 
+impl XvmCallOk {
+    pub fn output(&self) -> &Vec<u8> {
+        &self.output
+    }
+}
+
 /// Denotes an successful XVM call execution
 #[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, scale_info::TypeInfo)]
 pub struct XvmCallError {
@@ -88,6 +94,12 @@ pub struct XvmCallError {
     error: XvmError,
     /// Total consumed weight. This is in context of Substrate (1 unit of weight ~ 1 ps of execution time)
     consumed_weight: u64,
+}
+
+impl XvmCallError {
+    pub fn error(&self) -> &XvmError {
+        &self.error
+    }
 }
 
 /// Result for executing X-VM calls
--- a/frame/pallet-xvm/src/lib.rs
+++ b/frame/pallet-xvm/src/lib.rs
@@ -81,7 +81,7 @@ pub struct XvmCallOk {
 }
 
 impl XvmCallOk {
-    pub fn output(&self) -> &Vec<u8> {
+    pub fn output(&self) -> &[u8] {
         &self.output
     }
 }
--- a/frame/pallet-xvm/src/pallet/mod.rs
+++ b/frame/pallet-xvm/src/pallet/mod.rs
@@ -64,13 +64,11 @@ pub mod pallet {
 
     impl<T: Config> Pallet<T> {
         pub fn xvm_bare_call(
-            // origin: OriginFor<T>,
             context: XvmContext,
             from: T::AccountId,
             to: Vec<u8>,
             input: Vec<u8>,
         ) -> XvmResult {
-            // let from = ensure_signed(origin)?;
             let result = T::SyncVM::xvm_call(context, from, to, input);
 
             log::trace!(
--- a/frame/pallet-xvm/src/pallet/mod.rs
+++ b/frame/pallet-xvm/src/pallet/mod.rs
@@ -63,6 +63,12 @@ pub mod pallet {
     }
 
     impl<T: Config> Pallet<T> {
+        /// Internal interface for cross-pallet invocation.
+        /// Essentially does the same thing as `xvm_call`, but a bit differently:
+        ///   - It does not verify origin
+        ///   - It does not use `Dispatchable` API (cannot be called from tx)
+        ///   - It does not deposit event upon completion
+        ///   - It returns `XvmResult` letting the caller get return data directly
         pub fn xvm_bare_call(
             context: XvmContext,
             from: T::AccountId,
--- a/frame/pallet-xvm/src/pallet/mod.rs
+++ b/frame/pallet-xvm/src/pallet/mod.rs
@@ -62,6 +62,26 @@ pub mod pallet {
         XvmQuery { result: Result<Vec<u8>, XvmError> },
     }
 
+    impl<T: Config> Pallet<T> {
+        pub fn xvm_bare_call(
+            // origin: OriginFor<T>,
+            context: XvmContext,
+            from: T::AccountId,
+            to: Vec<u8>,
+            input: Vec<u8>,
+        ) -> XvmResult {
+            // let from = ensure_signed(origin)?;
+            let result = T::SyncVM::xvm_call(context, from, to, input);
+
+            log::trace!(
+                target: "xvm::pallet::xvm_bare_call",
+                "Execution result: {:?}", result
+            );
+
+            result
+        }
+    }
+
     #[pallet::call]
     impl<T: Config> Pallet<T> {
         #[pallet::call_index(0)]
--- a/frame/pallet-xvm/src/wasm.rs
+++ b/frame/pallet-xvm/src/wasm.rs
@@ -63,7 +63,7 @@ where
         })?;
         let call_result = pallet_contracts::Pallet::<T>::bare_call(
             from, // no need to check origin, we consider it signed here
-            dest, // no need to lookup since it's already `AccountId`
+            dest,
             Default::default(),
             gas_limit.into(),
             None,
@@ -83,7 +83,6 @@ where
             .ref_time();
 
         match call_result.result {
-            // FIXME What about `success.flags`?
             Ok(success) => Ok(XvmCallOk {
                 output: success.data,
                 consumed_weight,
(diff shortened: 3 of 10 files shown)


PierreOssun (approved):
LGTM

frame/pallet-xvm/src/lib.rs
```diff
@@ -80,6 +80,12 @@ pub struct XvmCallOk {
     consumed_weight: u64,
 }
 
+impl XvmCallOk {
+    pub fn output(&self) -> &Vec<u8> {
```
Dinonard: Just wondering, how come you're using direct Vec reference instead of slice?

I thought that in this context, `slice` would be more idiomatic.

precompiles/xvm/src/lib.rs
```diff
@@ -95,18 +95,40 @@ where
...
+                log::trace!(
+                    target: "xvm-precompile::xvm_call",
+                    "success: {:?}", success
+                );
+
+                Ok(succeed(
+                    EvmDataWriter::new()
+                        .write(true)
+                        .write(pallet_xvm::consumed_weight(&result))
+                        .write(Bytes(success.output().clone())) // FIXME redundant clone
```
Dinonard: Perhaps it's better to leave a **TODO** given we have those in this repository. We don't have any **FIXME**.

frame/pallet-xvm/src/wasm.rs
```diff
@@ -55,39 +56,42 @@ where
...
-        } else {
-            gas_limit.ref_time()
-        };
-        Ok(XvmCallOk {
-            output: Default::default(), // TODO: Fill in with output from the call
-            consumed_weight,
-        })
+        let consumed_weight = call_result
+            .gas_consumed
+            .saturating_add(T::WeightInfo::call())
```
Dinonard: Not sure why you need to add additional weight here?

The `bare_call` is used for dry-run so the value returned should be exactly what's actually consumed.

frame/pallet-xvm/src/pallet/mod.rs
```diff
@@ -62,6 +62,24 @@ pub mod pallet {
         XvmQuery { result: Result<Vec<u8>, XvmError> },
     }
 
+    impl<T: Config> Pallet<T> {
+        pub fn xvm_bare_call(
```
Dinonard, line 72: Docs

frame/pallet-xvm/src/lib.rs
0x7CFE: You're right that slice would be more idiomatic. 

Originally I did `&Vec` because currently we have `EvmData` implementation for `Vec` only and not for slice, so to use `output` in `EvmDataWriter` I need either a clone of vector or a reference. But then I found out that I need to use `Bytes(Vec<u8>)` to do compact byte encoding.

Anyways, good catch, will change it to slice.

frame/pallet-xvm/src/wasm.rs
0x7CFE: Originally the code, instead of `bare_call`, contained a call to [the pallet's `call` method](https://github.com/paritytech/substrate/blob/master/frame/contracts/src/lib.rs#L635) that internally does 

```rust
    output.gas_meter.into_dispatch_result(output.result, T::WeightInfo::call())
```

which, in turn does the following:
```rust
	/// Turn this GasMeter into a DispatchResult that contains the actually used gas.
	pub fn into_dispatch_result<R, E>(
		self,
		result: Result<R, E>,
		base_weight: Weight,
	) -> DispatchResultWithPostInfo
	where
		E: Into<ExecError>,
	{
		let post_info = PostDispatchInfo {
			actual_weight: Some(self.gas_consumed().saturating_add(base_weight)),
			pays_fee: Default::default(),
		};

		result
			.map(|_| post_info)
			.map_err(|e| DispatchErrorWithPostInfo { post_info, error: e.into().error })
	}
```

It could be that I misinterpreted the logic here.

precompiles/xvm/src/lib.rs
```diff
@@ -95,18 +95,40 @@ where
...
+            result @ Ok(success) => {
+                log::trace!(
+                    target: "xvm-precompile::xvm_call",
+                    "success: {:?}", success
+                );
+
+                Ok(succeed(
+                    EvmDataWriter::new()
+                        .write(true)
+                        .write(pallet_xvm::consumed_weight(&result))
```
akru: In my opinion it's good enough to return just `(bool, bytes)` tuple. Consumed weight is for charge gas on precompile side.

precompiles/xvm/src/lib.rs
```diff
@@ -95,18 +95,40 @@ where
...
+                    "failure: {:?}", failure
+                );
+
+                let mut error_buffer = Vec::new();
+                failure.error().encode_to(&mut error_buffer);
+
+                Ok(succeed(
+                    EvmDataWriter::new()
+                        .write(false)
+                        .write(pallet_xvm::consumed_weight(&result))
```
akru: Same here.

akru (approved):

frame/pallet-xvm/src/wasm.rs
0x7CFE: @akru what's your thoughts on this?

Dinonard: You should go through the code, and understand why that extra weight is added there, and why the _dry-run_ doesn't use it.

0x7CFE: This is not dry run. In pallet contracts, both `call` and `bare_call` use exactly the same `Self::internal_call`, but the former converts the result to dispatch result.

```rust
		#[pallet::call_index(6)]
		#[pallet::weight(T::WeightInfo::call().saturating_add(*gas_limit))]
		pub fn call(
			origin: OriginFor<T>,
			dest: AccountIdLookupOf<T>,
			#[pallet::compact] value: BalanceOf<T>,
			gas_limit: Weight,
			storage_deposit_limit: Option<<BalanceOf<T> as codec::HasCompact>::Type>,
			data: Vec<u8>,
			let gas_limit: Weight = gas_limit.into();
			let origin = ensure_signed(origin)?;
			let dest = T::Lookup::lookup(dest)?;
			let mut output = Self::internal_call(
				origin,
				dest,
				value,
				gas_limit,
				storage_deposit_limit.map(Into::into),
				data,
				None,
				Determinism::Deterministic,
			);
			if let Ok(retval) = &output.result {
				if retval.did_revert() {
					output.result = Err(<Error<T>>::ContractReverted.into());
				}
			}
			output.gas_meter.into_dispatch_result(output.result, T::WeightInfo::call())
		}
```

```rust
	pub fn bare_call(
		origin: T::AccountId,
		dest: T::AccountId,
		value: BalanceOf<T>,
		gas_limit: Weight,
		storage_deposit_limit: Option<BalanceOf<T>>,
		data: Vec<u8>,
		debug: bool,
		determinism: Determinism,
	) -> ContractExecResult<BalanceOf<T>> {
		let mut debug_message = if debug { Some(DebugBufferVec::<T>::default()) } else { None };
		let output = Self::internal_call(
			origin,
			dest,
			value,
			gas_limit,
			storage_deposit_limit,
			data,
			debug_message.as_mut(),
			determinism,
		);
		ContractExecResult {
			result: output.result.map_err(|r| r.error),
			gas_consumed: output.gas_meter.gas_consumed(),
			gas_required: output.gas_meter.gas_required(),
			storage_deposit: output.storage_deposit,
			debug_message: debug_message.unwrap_or_default().to_vec(),
		}
	}
```

As I mentioned before, I did that addition to keep the behavior exactly the same as it was before the change.

Dinonard: > This is not dry run. In pallet contracts, both call and bare_call use exactly the same Self::internal_call, but the former converts the result to dispatch result.

Please check how the `bare_call ` is used in `Astar` repo (runtime code).

> As I mentioned before, I did that addition to keep the behavior exactly the same as it was before the change.

And as I mentioned before so many times, we should understand WHY we're doing what we are doing.
Copy/pasting it from somewhere and saying that's the reason is not good enough.

This is the code from runtime code (for the dry-run):
```Rust
        fn call(
            origin: AccountId,
            dest: AccountId,
            value: Balance,
            gas_limit: Option<Weight>,
            storage_deposit_limit: Option<Balance>,
            input_data: Vec<u8>,
        ) -> pallet_contracts_primitives::ContractExecResult<Balance> {
            let gas_limit = gas_limit.unwrap_or(RuntimeBlockWeights::get().max_block);
            Contracts::bare_call(origin, dest, value, gas_limit, storage_deposit_limit, input_data, true, pallet_contracts::Determinism::Deterministic)
        }
```

Why isn't the extra weight added here as well?

PR #148, opened 2023-04-24
Pallet-scheduler Chain extension

PierreOssun:
**Pull Request Summary**
Added two calls for pallet-scheduler chain extension:
_Schedule_
It schedule a call to another smart contract. It can be periodic or not. Origin is set as contract address

_Cancel_
The caller can cancel its call (only smart-contracts can be caller)

Also added tests for it.

**Check list**
- [x] added unit tests


chain-extensions/pallet-scheduler/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+>>::Balance;
+
+/// Pallet Scheduler chain extension.
+pub struct SchedulerExtension<T, W>(PhantomData<(T, W)>);
+
+impl<T, W> Default for SchedulerExtension<T, W> {
+    fn default() -> Self {
+        SchedulerExtension(PhantomData)
+    }
+}
```
ashutoshvarma, line 75:
```suggestion
#[derive(DefaultNoBound)]
pub struct SchedulerExtension<T, W>(PhantomData<(T, W)>);
```

No need to manually impl `Default`, use `use frame_support::DefaultNoBound;`

chain-extensions/pallet-scheduler/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn try_from(value: u16) -> Result<Self, Self::Error> {
+        match value {
+            1 => Ok(SchedulerFunc::Schedule),
+            2 => Ok(SchedulerFunc::Cancel),
+            _ => Err(DispatchError::Other(
+                "PalletSchedulerExtension: Unimplemented func_id",
+            )),
+        }
+    }
+}
```
ashutoshvarma, line 62:
```suggestion
#[repr(u16)]
#[derive(TryFromPrimitive)]
enum SchedulerFunc {
    Schedule = 1,
    Cancel = 2,
}
```

Using `num_enum` will make this less error prone, right now enum values are not mapping 1 to 1 with `TryFrom` implementation which is not ideal.

chain-extensions/types/scheduler/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+pub enum Origin {
+    Caller,
+    Address,
+}
+
+impl Default for Origin {
+    fn default() -> Self {
+        Self::Address
+    }
+}
```
ashutoshvarma, line 85:
```suggestion
#[derive(Default, Debug, Copy, Clone, PartialEq, Eq, Encode, Decode, MaxEncodedLen)]
#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
pub enum Origin {
    Caller,
    #[default]
    Address,
}
```

ashutoshvarma (changes requested):

ashutoshvarma:
A General comment on the CE code structure we follow currently.

We have `types` crate for each CE to allow importing CE types without depending on CE itself (at least I believe that, correct me if wrong). That is useful for ink! contracts to build the CE call as they required the input, output and error types that are being used in CE.

If above assumption is correct then why is there a need to copy past the types to contract helpers in `chain-extension-contract` repo - https://github.com/swanky-dapps/chain-extension-contracts/blob/ad0787b7e0e087493efaf00610a027356603a0a9/crates/scheduler/lib.rs#L44-L112?
Why not simply add `types` as dep to contract and import it directly? Maybe I'm missing something?

PierreOssun:
Because of the features hell that it will require:
Check this [discussion](https://github.com/AstarNetwork/astar-frame/pull/132#discussion_r1113135744)

ashutoshvarma:
I understand, we don't want mutually exclusive features to make sure `cargo check --all-features` always work.

From my understanding, in the `types` crate we only need `ink_env` dependency because we need to implement `FromStatusCode` trait for the `Error`/`Outcome` type. Since clearly, we cannot have `ink_env` as dep in `types` (it depends on `ink_allocator` which cannot be built in `std`) so `FromStatusCode` then needs to be implemented in contract side which we cannot as implementing external traits for external types in not allowed in rust.

We have few options to fix this,

1. Implement a decl macro which will generate the `Outcome`/`Error` enum. This can then be used in contract side to generate the same enum.

2. Wrap the `Outcome`/`Error` in struct in contract side and implement the  `FromStatusCode` on it. Since scale encoding for wrapper is exactly the same to the inner types, we don't have to worry about serialization clash.

3. Open issue in substrate to move `ink_env::chain_extension::FromStatusCode` to `ink_primitives::chain_extension::FromStatusCode`

ashutoshvarma:
Here is the example for 1st approach, (code from XCM CE)
The `primitives` crate has all the shared types which both pallet/CE and ink! contract can share.
This is the macro to generate CE `Error` type - https://github.com/AstarNetwork/Astar/blob/8441e0f3715f2389c8754c74a6a50500d204525c/xcm/pallet-xcm-transactor/primitives/src/lib.rs#L72-L99 
and it is used like this in ink! contract - https://github.com/AstarNetwork/Astar/blob/8441e0f3715f2389c8754c74a6a50500d204525c/xcm/pallet-xcm-transactor/contracts-example/sdk.rs#L9-L18

PierreOssun:
@ashutoshvarma @Dinonard I think we can close this

PR #155, opened 2023-05-11
dApps staking v3 - part 1

Dinonard:
**Pull Request Summary**

First PR that will introduce part of dApps staking v3 functionality.
Overall implementation will be broken down into multiple PRs to make it easier for reviewers.

**Part 1**
* maintenance mode
* basic operations for registering dApps
* dApp unregistration, reward destination & owner manipulation
* various struct & types to simplify the implementation
* balance lock implementation

After receiving first review, will amend the comments and implement extensive unit tests.

**Check list**
- [x] dApp manipulation & balance lock implemented 
- [x] first review received
- [x] added unit tests


frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,693 @@
...
+                        Error::<T>::NotOperatedDApp
+                    );
+
+                    dapp_info.state = DAppState::Unregistered(current_era);
+
+                    Ok(().into())
+                },
+            )?;
+
+            // TODO: might require some modification later on, like additional checks to ensure contract can be unregistered.
```
gitofdeepanshu, line 360: what are the scenarios in which contract can't be unregistered?

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,693 @@
...
+}
+
+// TODO: would users get better UX if we kept using eras? Using blocks is more precise though.
+/// How much was unlocked in some block.
+#[derive(Encode, Decode, MaxEncodedLen, Clone, Copy, Debug, PartialEq, Eq, TypeInfo)]
+pub struct UnlockingChunk<Balance: AtLeast32BitUnsigned + MaxEncodedLen + Copy> {
+    #[codec(compact)]
+    amount: Balance,
+    #[codec(compact)]
+    unlock_block: BlockNumber,
```
gitofdeepanshu: `BlockNumber` is undefined here.

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,693 @@
...
+
+pub use pallet::*;
+
+const STAKING_ID: LockIdentifier = *b"dapstake";
+
+/// The balance type used by the currency system.
+pub type BalanceOf<T> =
+    <<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
+
+/// Concenience type for `AccountLedger` usage.
```
gitofdeepanshu: is this a typo? `convenient` maybe?

frame/dapp-staking-v3/src/lib.rs
Dinonard: What should be avoided is unregistering a contract while leaving storage items behind, _forever_.
So later on, when PR progresses more, this TODO is a reminder to revisit the issue (I don't have a clear solution for this yet).

frame/dapp-staking-v3/src/lib.rs
Dinonard: Right, late minute breaking changes 😂

frame/dapp-staking-v3/src/lib.rs
Dinonard: Yes, should be convenience.

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,730 @@
...
+use frame_system::pallet_prelude::*;
+use parity_scale_codec::{Decode, Encode};
+use sp_runtime::traits::{AtLeast32BitUnsigned, BadOrigin, Saturating, Zero};
+
+pub use pallet::*;
+
+#[cfg(test)]
+mod test;
+
+const STAKING_ID: LockIdentifier = *b"dapstake";
```
shaunxw: I wonder what's the migration plan to v3? If v3 coexists with prev version, probably best to use a different lock ID.

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,730 @@
...
+pub type PeriodNumber = u32;
+/// Dapp Id type
+pub type DAppId = u16;
+
+/// Distinct period types in dApp staking protocol.
+#[derive(Encode, Decode, MaxEncodedLen, Clone, Copy, Debug, PartialEq, Eq, TypeInfo)]
+pub enum PeriodType {
+    /// Period during which the focus is on voting.
+    /// Inner value is the era in which the voting period ends.
+    VotingPeriod(#[codec(compact)] EraNumber),
```
shaunxw:
```suggestion
    Voting {
        #[codec(compact)]
        ends_at: EraNumber,
    }),
```

It is more about personal preference, but I think this is more readable and self-explanatory. `PeriodType` naming includes `Period` space so we don't need to repeat.

If it makes sense, we can apply this to other enum fields.

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,730 @@
...
+        } else {
+            locked_chunk.era = era;
+            self.locked.try_push(locked_chunk).map_err(|_| ())?;
+        }
+
+        Ok(())
+    }
+}
+
+/// Rewards pool for lock participants  & dApps
```
shaunxw:
```suggestion
/// Rewards pool for lock participants & dApps
```

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,730 @@
...
+    #[pallet::generate_deposit(pub(crate) fn deposit_event)]
+    pub enum Event<T: Config> {
+        /// A smart contract has been registered for dApp staking
+        DAppRegistered {
+            owner: T::AccountId,
+            smart_contract: T::SmartContract,
+            dapp_id: DAppId,
+        },
+        /// dApp reward destination has been updated.
+        DAppRewardDestination {
```
shaunxw:
```suggestion
        DAppRewardDestinationUpdated {
```

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,730 @@
...
+            ensure!(!amount_to_lock.is_zero(), Error::<T>::ZeroAmount);
+
+            // Only lock for the next era onwards.
+            let lock_era = state.era.saturating_add(1);
+            ledger
+                .add_lock_amount(amount_to_lock, lock_era)
+                .map_err(|_| Error::<T>::TooManyLockedBalanceChunks)?;
+            ensure!(
+                ledger.locked_amount().saturating_add(amount_to_lock)
+                    > T::MinimumLockedAmount::get(),
```
shaunxw:
```suggestion
                    >= T::MinimumLockedAmount::get(),
```

A minimum amout should be valid right?

shaunxw (commented):
Looks good in general!

I wonder why we are using `DispatchResultWithPostInfo` as the return type though there is no post info?

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,730 @@
...
+}
+
+/// General information about dApp.
+#[derive(Encode, Decode, MaxEncodedLen, Clone, Copy, Debug, PartialEq, Eq, TypeInfo)]
+pub struct DAppInfo<AccountId> {
+    /// Owner of the dApp, default reward beneficiary.
+    pub owner: AccountId,
+    /// dApp's unique identifier in dApp staking.
+    #[codec(compact)]
+    pub id: DAppId,
```
shaunxw: Forgot to add this one: do we need `id` when they can be identified via `T::SmartContract`?

frame/dapp-staking-v3/src/lib.rs
Dinonard: v3 will replace the old version, and we want to keep the existing locks.

frame/dapp-staking-v3/src/lib.rs
Dinonard: I agree 😁 

Haven't gotten to the part of implementation where I use this, but I'd probably get the same idea as you after writing down `PeriodType::VotingPeriod` xD

frame/dapp-staking-v3/src/lib.rs
Dinonard: Right, also discovered this via UT 🙂

frame/dapp-staking-v3/src/lib.rs
Dinonard: ID was added because of how dApp rewards will be calculated.
Using `T::SmartContract` as identifier works, logically but it takes a lot of space.
To optimize this, a more compact _secundary_ (not sure if that's the right term) is introduced.

Right now it's part of the design (more info in the doc) but it might change in the future.

Dinonard:
> Looks good in general!
> 
> I wonder why we are using `DispatchResultWithPostInfo` as the return type though there is no post info?

Thanks!

For the `DispatchResultWithPostInfo` you're right that it's not needed. I just copied it from the previous pallet for the first extrinsic I added, then reused it afterwards 😁. Fixed it now.

Dinonard:
All functionality intended for this PR has been implemented, and it's ready for final review 🙏

frame/dapp-staking-v3/src/lib.rs
shaunxw: I see. Then it makes sense to use the existing lock ID.

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        /// Pallet is disabled/in maintenance mode.
+        Disabled,
+        /// Smart contract already exists within dApp staking protocol.
+        ContractAlreadyExists,
+        /// Maximum number of smart contracts has been reached.
+        ExcededMaxNumberOfContracts,
```
shaunxw:
```suggestion
        ExceededMaxNumberOfContracts,
```

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+
+            NextDAppId::<T>::put(dapp_id.saturating_add(1));
+
+            Self::deposit_event(Event::<T>::DAppRegistered {
+                owner,
+                smart_contract,
+                dapp_id,
+            });
+
+            Ok(().into())
```
shaunxw:
```suggestion
            Ok(())
```

`into` isn't needed anymore with `DispatchResult` as return type. Same to other calls.

frame/dapp-staking-v3/src/test/mock.rs
```diff
@@ -0,0 +1,182 @@
...
+///
+/// Function has no effect if era is already passed.
+pub(crate) fn advance_to_era(era: EraNumber) {
+    // TODO: Properly implement this later when additional logic has been implemented
+    ActiveProtocolState::<Test>::mutate(|state| state.era = era);
+}
+
+/// Initialize first block.
+/// This method should only be called once in a UT otherwise the first block will get initialized multiple times.
+pub(crate) fn initialize_first_block() {
```
shaunxw: Could we have it executed in `build` function, so it doesn't have to repeat in each UT?

frame/dapp-staking-v3/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+        Locked {
+            account: T::AccountId,
+            amount: BalanceOf<T>,
+        },
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        /// Pallet is disabled/in maintenance mode.
+        Disabled,
```
shaunxw, line 138: IMO disabled/maintainance mode should also be covered in unit tests.

shaunxw (commented):
Looks good overall! Just some small issues.

frame/dapp-staking-v3/src/lib.rs
Dinonard: Right, I rushed! Will clean it up!

frame/dapp-staking-v3/src/test/mock.rs
Dinonard: Sure, good suggestion! Nice way to get `before_each` hook 🙂

frame/dapp-staking-v3/src/lib.rs
Dinonard: True, I didn't forget about it though, just thought I'd cover it at the very end.
Personal bias since for `v2`, I added maintenance mode only later.

Will add it now though!

shaunxw (approved):

Dinonard:
@shaunxw thanks for the approval!

I was thinking, instead of merging it directly, to keep this in a 'live' branch.
All consecutive PRs would be merged to this branch instead (or I could create another branch to aggregate all PRs).

This would keep the review of PRs light(er), while also providing one big overview in the end, before everything is merged.
What do you think?

shaunxw:
Love this idea. It definitely helps with the review process.

